PR #16, opened 2022-06-19
feat: 404 page added

sumana2001:
Screenshot 2022-06-20 at 12 59 30 AM


src/pages/not-found/index.jsx
```diff
@@ -0,0 +1,48 @@
+import * as React from "react";
+import Box from "@mui/material/Box";
+import { useTranslation } from "react-i18next";
+
+export default function Insights() {
+  const { t } = useTranslation();
+  return (
+    <center>
```
alexfauquette: I was not aware of the existence of this element. It seems to be [deprecated](https://developer.mozilla.org/fr/docs/Web/HTML/Element/center)

`<Box sx={{textAlign: 'center'}}` is the recommended way to get similar result

src/pages/not-found/index.jsx
```diff
@@ -0,0 +1,48 @@
...
+      <Box
+        component="img"
+        sx={{
+          width: { xs: "100%", md: "60%" },
+          height: "auto",
+        }}
+        alt="404 not found"
+        src={require("../../assets/404.png")}
+      />
+      <Box
```
alexfauquette: Boxes are a convenient way to structure your page with `div`. But here you are adding text. So in HTML it should be `<p>`.

You have a convenient way to do that in MUI [Typography](https://mui.com/material-ui/react-typography/#main-content) In which you play mainly with 2 props: `variant` for the styling and `component` for the web sementic.

src/pages/not-found/index.jsx
```diff
@@ -0,0 +1,48 @@
...
+        sx={{
+          typography: "h1",
+          fontSize: "2em",
+          marginTop: "-5rem",
```
alexfauquette: Using a negative margin is not recommended 😉 

image

alexfauquette (approved):

alexfauquette:
Nice looking page. I added some feedback, and since I was trying them during the review I pushed a commit with the fixes. If you're okay with the modification, we can merge

PR #17, opened 2022-06-21
feat: add question history component

alexfauquette:
### What
- the buffer reducer is modified such as it keeps track of how many questions remain to answer (ignoring skipped ones)
- Display the X lest questions annotated with either yes or no

image


alexfauquette:
@sumana2001 That will probably interest you for the responsive design

If it's ok for you you can merge it. Don't hesitate to add comment if you see things to modify

sumana2001:
Looks good to me!

PR #51, opened 2022-07-10
anticipate image download for next questions

alexfauquette:
When we display the question of index 0 in the buffer, we also ask the browser to fetch images with indexes 1, 2, 3, 4

such that time between two question display is reduced


CharlesNepote (approved):
Looks good to me, but I'm not a React expert. It might also be reviewed by a REACT expert.

PR #54, opened 2022-07-15
Shmigoll design issues

Shmigolk:
### What
- description of the PR
Just made this: 
image


teolemon:
4:12:25 PM: src/pages/questions/QuestionDisplay.jsx
4:12:25 PM:   Line 86:6:  React Hook React.useEffect has a missing dependency: 'answerQuestion'. Either include it or remove the dependency array. If 'answerQuestion' changes too often, find the parent component that defines it and wrap that definition in useCallback  react-hooks/exhaustive-deps

Issue #37, opened 2022-07-04
What can I work on ?


teolemon:
#### Logos game
- [ ] Make multi-select more intuitive with an optional checkbox
- [ ] Add a prompt explaining how to play: select all similar logos, and then enter their name and type on top, and click save
#### Eco-Score
- [ ] Add a little sentence on top: you can help with the eco-score
#### Insight view
- [ ] Swap icons
  - [ ] <img width="157" alt="image" src="https://user-images.githubusercontent.com/1689815/177147271-8ade02bd-a1b4-4677-a7a2-f77d8627b567.png">
- [ ] Make the barcodes clickable (either link to product page, or link to all insights with that barcode, or both)
- [ ] Make capitalization consistent
  - [ ] <img width="157" alt="image" src="https://user-images.githubusercontent.com/1689815/177147339-ca32f2d2-823a-460a-9dec-7b0afbe2577d.png">
#### Category/Brands/… game
- [ ] Images of the products should be aligned, possibly in a grid, possibly minimized ?
  - [ ] <img width="426" alt="image" src="https://user-images.githubusercontent.com/1689815/177146401-789f7b57-2eda-4872-8d58-866086d3f0a4.png">
- [ ] The Hunger Games logo is not aligned with the menus in the top bar
  - [ ] <img width="706" alt="image" src="https://user-images.githubusercontent.com/1689815/177146513-265ea486-6240-4f3a-9e50-86e51b2d7ba5.png">
- [ ] Add padding after "See examples of this brand"
  - [ ] <img width="347" alt="image" src="https://user-images.githubusercontent.com/1689815/177146573-788cd6f0-607a-4694-8b5c-22da329c8380.png">
- [ ] View and Edit button should be augmented with icons, and possibly be turned into buttons
  - [ ] <img width="347" alt="image" src="https://user-images.githubusercontent.com/1689815/177146705-29b2bb57-8f79-4c45-970f-ec03176fa9e7.png">
- [ ] brand / category / label dropdown should be more explicit "Only show brands" "Only show categories"
- [ ] Buttons should be centered and separated by a few horizontal pixels in mobile mode
  - [ ] 
- [ ] View should be centered in mobile mode

#### Settings

- [ ] Add a Dev mode
- [ ] Hide the Insights view by default
- [ ] Add an option in dev mode to show insights view
- [ ] Add an option to see all secondary / hidden views (even on a legacy install of Hunger Games)

### Navigation

- [ ] Use the full width of the page
[...]

PR #56, opened 2022-07-17
No questions remaining UI

sumana2001:
### What
- We should prompt users to contribute more. One of the ways is to redirect our audience from hunger games to the app so that we can get more and more data. 
- When there are no questions remaining, the screen will ask the user to install the app. 

### Screenshot
Screenshot 2022-07-17 at 6 37 03 PM

### Part of 
- #37


src/pages/questions/QuestionDisplay.jsx
```diff
@@ -57,13 +57,67 @@ const getValueTagExamplesURL = (question) => {
...
+              alt="Appstore"
+              width="140px"
+              height="60px"
+            />
+          </a>
+        </Stack>
+        <Stack direction="row" spacing={1}>
+          <a href="https://www.microsoft.com/en-us/p/openfoodfacts/9nblggh0dkqr">
+            <img
+              src="	https://world.openfoodfacts.org/images/misc/microsoft/English.svg"
```
alexfauquette, line 104:
```suggestion
              src="https://world.openfoodfacts.org/images/misc/microsoft/English.svg"
```

src/pages/questions/QuestionDisplay.jsx
```diff
@@ -57,13 +57,67 @@ const getValueTagExamplesURL = (question) => {
...
+          direction={{ xs: "column", sm: "row" }}
+          alignItems="center"
+          spacing={{ xs: 0, sm: 2 }}
+        >
+          <h2>{t("questions.no_questions_remaining")}</h2>
+          <Button size="small" variant="contained" onClick={resetFilters}>
+            Reset filters
+          </Button>
+        </Stack>
+        <h1>OR</h1>
```
alexfauquette, line 78: `<h1>` has a meaning: this is the top title of the page. I assume the top title of this page is not "OR" 😉

Same for `h2`, `h3`, ... If you want a big text use CSS, not HTML

On this topic, you can read [`<Typography/>` doc](https://mui.com/material-ui/react-typography/#main-content)

```jsx
<Typography variant="h1" component="p">A big text</Typography>
```

This components hase two important props:
- `variant`: defines how the text will look like (here it will look like a top title)
- `component`the HML compont which will be used (the sementic meaning (here it's a basic text)

sumana2001: Oh okay, got it. Thank you. I'll make the changes

src/pages/questions/index.jsx
```diff
@@ -18,6 +18,15 @@ export default function Questions() {
     useQuestionBuffer(filterState);
   const question = buffer[0] ?? null;
 
+  const resetFilters = React.useCallback(
```
nobeeakon, line 21: useCallback (memoization) may not be needed in this case, maybe go with a normal arrow function?

alexfauquette: The `setFilterState` is not usefull in the array of dependencies, because state setter are guaranteed to stay constant between render

It's not entirely useless, because `resetFilters` is passed to `QuestionDisplay`. WIthout that useCallback each render will create a new function, and so `<QuestionDisplay/>` will rerender too

nobeeakon: Yeah, I guess that the baseline is if the useCallback memoization improves performance. I think that in this case as the QuestionDisplay is a presentational component and doesn't have heavy computations or something really slow, there is not a lot of benefit in memoizing the reset function and makes the code a bit more complex. Also, the memoization is not cost free, it will store in memory and every render needs to compare with dependency array. Besides that, the QuestionDisplay will rerender every time the Questions (index) rerenders, unless QuestionDisplay itself is memoized.

PR #60, opened 2022-07-18
Shmigol issues

Shmigolk:
### What
align pictures 
image

align btns 
image

fix issue about keyboards shortcut by adding eventListener to document with useEffect.

### DIFF
--- a/src/pages/questions/QuestionDisplay.jsx
+++ b/src/pages/questions/QuestionDisplay.jsx
@@ -115,8 +115,9 @@ const QuestionDisplay = ({ question, answerQuestion }) => {
       </Box>
       <Stack
         direction="row"
-        justifyContent="space-around"
+        justifyContent="center"
         flexWrap="wrap"
+        gap={'.9em'}
         onKeyDown={(event) => {
           switch (event.key) {
             case "k":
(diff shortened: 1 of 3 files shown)


teolemon:
4:12:25 PM:   Line 86:6:  React Hook React.useEffect has a missing dependency: 'answerQuestion'. Either include it or remove the dependency array. If 'answerQuestion' changes too often, find the parent component that defines it and wrap that definition in useCallback  react-hooks/exhaustive-deps

src/pages/questions/QuestionDisplay.jsx
```diff
@@ -58,17 +57,49 @@ const getValueTagExamplesURL = (question) => {
...
-const QuestionDisplay = ({ question, answerQuestion }) => {
+const QuestionDisplay = ({ question, answerQuestion, resetFilters }) => {
   const { t } = useTranslation();
   const valueTagQuestionsURL = getValueTagQuestionsURL(question);
   const valueTagExamplesURL = getValueTagExamplesURL(question);
 
+  React.useEffect(() => {
+
+    function handleShortCut(event) {
+      if (question) {
```
alexfauquette, line 68: You should also verify if the event comes from an `inputè element or not

Here is an example in which I press K in an input, so it adds a "k" to the input value (the expected behavior) but it also skip the question (not expected behavior)



https://user-images.githubusercontent.com/45398769/180067577-f36ab90c-35a9-4a22-a29c-8cf80db05372.mp4

Here is a potential solution

https://stackoverflow.com/questions/7230976/javascript-event-handler-on-body-but-not-on-input

PR #67, opened 2022-07-21
Short cut issue solving

Shmigolk:
Solve the problem with shortcut working inside input areas.
Use event.keyCode instead of event.code


Shmigolk:
I think it is not the best solution. So, to prevent shortcut work during filling input I add check "If event.target is not an input element". But I think it is better maybe to use another solution, such as adding boolean state or to prevent shortcut working when we press EDIT FILTERS button. But we render buttons (skip, yes, no) when we editing filters that is why made a desision to leave the ability to use shortcut

PR #78, opened 2022-07-27
Fixed shortCut issue and added small UI improvements

Shmigolk:
Aligned pictures in Product information and changed navBar

image

### DIFF
--- a/src/components/ResponsiveAppBar.jsx
+++ b/src/components/ResponsiveAppBar.jsx
@@ -27,7 +27,7 @@ const pages = [
   { url: "eco-score", translationKey: "menu.eco-score" },
   { translationKey: "menu.manage" },
   { url: "insights", translationKey: "menu.insights" },
-  { url: "settings", translationKey: "menu.settings" },
+  // { url: "settings", translationKey: "menu.settings" },
 ];
 
 const ResponsiveAppBar = () => {
@@ -135,11 +135,11 @@ const ResponsiveAppBar = () => {
             sx={{
               display: { xs: "none", md: "flex" },
               flexDirection: "row",
-              alignItems: "center",
+              alignItems: "baseline",
             }}
           >
             <MuiLink
-              sx={{ mr: 1, display: "flex" }}
+              sx={{ mr: 1, display: "flex", alignSelf: 'center' }}
               href="https://world.openfoodfacts.org/"
               target="_blank"
             >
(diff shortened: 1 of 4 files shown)


src/components/ResponsiveAppBar.jsx
```diff
@@ -126,10 +127,19 @@ const ResponsiveAppBar = () => {
               display: { xs: "none", md: "flex" },
               flexDirection: "row",
               alignItems: "center",
+              width: '100%',
+              justifyContent: 'space-between'
```
alexfauquette, line 131: A hack to add some space is to set an empty `div` with flexGrowth = 1
```jsx
<Box sx={{flexGrowth: 1}}></Box>
```

This div will extend itself to fill all the remaining space

src/pages/questions/QuestionDisplay.jsx
```diff
@@ -125,7 +149,7 @@ const QuestionDisplay = ({ question, answerQuestion, resetFilters }) => {
         justifyContent="center"
         spacing={2}
         sx={{ mb: 1 }}
-        onKeyDown={(event) => {
+        /*onKeyDown={(event) => {
```
alexfauquette, line 152: You can delete it now

src/pages/questions/ProductInformation.jsx
```diff
@@ -122,7 +123,7 @@ const ProductInformation = ({ question }) => {
         labelPlacement="end"
       />
       {!hideImages && productData?.images && (
-        <Stack spacing={2} direction="row" flexWrap="wrap">
+        <Stack direction="row" flexWrap="wrap" gap='1em'>
```
nobeeakon, line 126: Possibly better to use [Grid](https://mui.com/material-ui/react-grid/)?

src/pages/questions/QuestionDisplay.jsx
```diff
@@ -62,6 +62,30 @@ const QuestionDisplay = ({ question, answerQuestion, resetFilters }) => {
...
+            answerQuestion({ value: 0, insightId: question.insight_id });
+            break;
+          default:
+            break;
+        }
+      }}
+
+    window.addEventListener( 'keydown', handleShortCut)
+    return () =>  window.removeEventListener('keydown', handleShortCut)
+  }, [question, answerQuestion])
```
nobeeakon, line 87: better to use question?.insight_id in the dependency array (as it is a string and  question is an object so will change more frequently). Also you may need to check if the value is not null at the begnning of the useEffect.

src/pages/questions/ProductInformation.jsx
alexfauquette: The idea was to fix image's width and let them wrap when needed, but using Grid with a number of images per row varying with breaking points sounds interesting

alexfauquette: I merged because the main topic was fixing the shortcuts. I open an issue for using Grid

Issue #59, opened 2022-07-18
App should memorize "Hide images" check or uncheck


CharlesNepote:
Step to reproduce:
* load a page
* uncheck "Hide images"
* do anything
* reload the page
* the "Hide images" is now checked

PR #80, opened 2022-07-27
issue 59, store hide images state

nobeeakon:
### What
- Hide images state now stored in the local storage as part of the settings

### Fixes bug(s)
- #59

### DIFF
--- a/src/localeStorageManager.js
+++ b/src/localeStorageManager.js
@@ -2,9 +2,22 @@ import isEqual from "lodash.isequal";
 
 const STORAGE_KEY = "hunger-game-settings";
 
+export const localSettingsKeys = {
+  language: 'lang',
+  isDevMode: 'devMode',
+  hideImages: 'questions_hideImages',
+}
+
+
 export const localSettings = {
   fetch: function () {
-    return JSON.parse(localStorage.getItem(STORAGE_KEY) || "{}");
+    let storedValue = {};
+    
+    try{
+      storedValue = JSON.parse(localStorage.getItem(STORAGE_KEY) || "{}");
+    } catch(e){}
+
+    return (typeof storedValue === 'object')?storedValue:{};
   },
   save: function (settings) {
     localStorage.setItem(STORAGE_KEY, JSON.stringify(settings));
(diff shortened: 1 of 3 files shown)


src/localeStorageManager.js
```diff
@@ -2,9 +2,22 @@ import isEqual from "lodash.isequal";
...
+  hideImages: 'questions_hideImages',
+}
+
+
 export const localSettings = {
   fetch: function () {
-    return JSON.parse(localStorage.getItem(STORAGE_KEY) || "{}");
+    let storedValue = {};
+    
+    try{
```
alexfauquette, line 16: In which case could it faill?

src/localeStorageManager.js
```diff
@@ -18,7 +31,13 @@ export const localSettings = {
 
 export const getIsDevMode = () => {
   const settings = localSettings.fetch();
-  return settings.devMode ?? false;
+  return settings[localSettingsKeys.isDevMode] ?? true;
```
alexfauquette, line 34: If the mode is not registered, we assume the value is `false` otherwise everyone would be in dev mode during their first visit

```suggestion
  return settings[localSettingsKeys.isDevMode] ?? false;
```

src/localeStorageManager.js
```diff
@@ -2,9 +2,22 @@ import isEqual from "lodash.isequal";
 
 const STORAGE_KEY = "hunger-game-settings";
 
+export const localSettingsKeys = {
+  language: 'lang',
+  isDevMode: 'devMode',
+  hideImages: 'questions_hideImages',
+}
```
alexfauquette, line 9: Love that 👍

src/pages/questions/ProductInformation.jsx
```diff
@@ -75,6 +77,11 @@ const ProductInformation = ({ question }) => {
     };
   }, [question?.barcode]);
 
+  const handleHideImages = (event) => {
+    setHideImages(event.target.checked);
+    localSettings.update(localSettingsKeys.hideImages, event.target.checked);
+  }
```
alexfauquette, line 83: Not for this PR, but it let me think that a hook `useSettingState` which is similar to `React.useState`+save data in settings could be interesting

alexfauquette (commented):
Nice, I just have a doubt about the default value of `getIsDevMode`

src/localeStorageManager.js
nobeeakon: the issue is that as you can  mess with the local storage it can have something that raises an error because is invalid, like the string "{a:[22"

src/localeStorageManager.js
nobeeakon: My bad, fixed now

src/pages/questions/ProductInformation.jsx
nobeeakon: yeah, that would be useful

PR #94, opened 2022-07-31
Show how many questions for a given value

### LINKED ISSUE #92: Getting the opportunity count for values in the question game
### What
image

- [ ] Getting the opportunity count for values in the question game
- [ ] Add an option in settings to open in new tab

### Why
- To know if a brand is worth clicking on

alexfauquette:
Fix #92

Add a Badge on top of the question value to display how many questions are associated to it.

Do not display it when the value is already specified in the filter, because it's not anymore an opportunity

image

### DIFF
--- a/src/pages/questions/QuestionDisplay.jsx
+++ b/src/pages/questions/QuestionDisplay.jsx
@@ -2,6 +2,7 @@ import * as React from "react";
 import { Link } from "react-router-dom";
 
 import Button from "@mui/material/Button";
+import Badge from "@mui/material/Badge";
 import Box from "@mui/material/Box";
 import Divider from "@mui/material/Divider";
 import Stack from "@mui/material/Stack";
@@ -17,6 +18,7 @@ import "react-medium-image-zoom/dist/styles.css";
 import { useTranslation } from "react-i18next";
 import { NO_QUESTION_LEFT, OFF_URL } from "../../const";
 import { reformatValueTag } from "../../utils";
+import robotoff from "../../robotoff";
 
 // const getFullSizeImage = (src) => {
 //   if (!src) {
@@ -43,6 +45,7 @@ const getValueTagQuestionsURL = (question) => {
   }
   return null;
 };
+
 const getValueTagExamplesURL = (question) => {
   if (
     question !== null &&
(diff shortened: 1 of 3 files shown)


src/pages/questions/QuestionDisplay.jsx
```diff
@@ -57,16 +60,55 @@ const getValueTagExamplesURL = (question) => {
...
   const valueTagExamplesURL = getValueTagExamplesURL(question);
+  const [nbOfPotentialQuestion, setNbOfPotentialQuestions] =
+    React.useState(null);
 
   React.useEffect(() => {
+    if (filterState.valueTag) {
+      // value is already in the filter so it's a useless information
+      setNbOfPotentialQuestions(null);
+      return;
+    } else {
```
nobeeakon, line 92: no need of the else as there is a return, removing the else also reduces code nesting so easier to read

src/pages/questions/QuestionDisplay.jsx
```diff
@@ -57,16 +60,55 @@ const getValueTagExamplesURL = (question) => {
...
+      let validRequest = true;
+      getNbOfQuestionForValue({
+        type: question?.insight_type,
+        valueTag: question?.value_tag,
```
nobeeakon, line 96: maybe add a check if those two values exist after the filterstate.value check?

src/pages/questions/QuestionDisplay.jsx
```diff
@@ -57,16 +60,55 @@ const getValueTagExamplesURL = (question) => {
...
+      }).then((nbQuestions) => {
+        if (validRequest) {
+          setNbOfPotentialQuestions(nbQuestions);
+        }
+      });
```
nobeeakon, line 101: maybe add a catch() {}

src/pages/questions/QuestionDisplay.jsx
```diff
@@ -57,16 +60,55 @@ const getValueTagExamplesURL = (question) => {
...
 
+  React.useEffect(() => {
     function handleShortCut(event) {
-      const preventShortCut = event.target.tagName.toUpperCase() === 'INPUT'
-      if (question && !preventShortCut) {
+      const preventShortCut = event.target.tagName.toUpperCase() === "INPUT";
+      if (question?.insight_id && !preventShortCut) {
         switch (event.keyCode) {
           case 75:
             answerQuestion({ value: -1, insightId: question.insight_id });
```
nobeeakon, line 114: off topic, but, would be nice to have somewhere an idea of what these values (-1, etc) represent, as just looking at the code one is kind of clueless

src/utils.js
```diff
@@ -32,7 +32,7 @@ export const reformatValueTag = (value) => {
 
 export const removeEmptyKeys = (obj) => {
   Object.keys(obj).forEach(
-    (key) => (obj[key] === undefined || obj[key] === "") && delete obj[key]
+    (key) => (obj[key] == null || obj[key] === "") && delete obj[key]
```
nobeeakon, line 35: I would advice against using == as having to remember transformation equivalences is really something, better to have something explicit as    obj[key] === null || obj[key] === undefined

alexfauquette: Of course, we can explicit every thing but `== null` is a short cut I use a lot when `undefined` or `nul should have same effect`

src/pages/questions/QuestionDisplay.jsx
alexfauquette: If a network error occurs, what should we do? I don't get the use of the `catch` here

nobeeakon: yeah, mmmm, the issue I'm thinking is that if an error accurs in the promise it will cause (from what I understand) the whole app to fail (unless there's  an error handler somewhere in the tree). I don't know what would work here, an option would be to have an error state and show it in the UI, but I guess that even just logging the error or doing nothing ... what do you think @alexfauquette ?

teolemon (approved):

PR #98, opened 2022-07-31
Welcome tour and Home page UI

sumana2001:
- Added the home page UI again as the other PR was having some errors in rendering
- Will add the welcome tour in this PR only


src/pages/home/homeCards.jsx
```diff
@@ -0,0 +1,75 @@
...
+    <Stack
+      spacing={3}
+      direction={{ xs: "column", sm: "column", md: "row" }}
+      alignItems="center"
+      justifyContent="center"
+      sx={{
+        marginBottom: "30px",
+      }}
+    >
+      {content.map((props) => (
```
nobeeakon, line 45: nit: by convention, props refer to the info passed to the component, having a variable that is called that way but it is not the passed props may be confusing. Maybe rename to something more explicit, like: content -> cards , and props to cardInfo

src/App.js
```diff
@@ -54,7 +54,6 @@ export default function App() {
           <CssBaseline />
           <ResponsiveAppBar />
           <Routes>
-            {/* TODO: put a home page for root url */}
```
nobeeakon, line 57: if not done in here, it would be nice to have an issue, and associate the TODO comment with the issue number, this way we don't forget about it :)

src/pages/home/homeCards.jsx
```diff
@@ -0,0 +1,75 @@
...
+        <Card sx={{ width: 350, height: 300 }} key={props.title}>
+        <CardActionArea component={Link} to={props.link}>     
+            <CardMedia
+              component="img"
+              height="200"
+              image={props.image}
+              alt=""
```
nobeeakon, line 52: if possible, please include the alt

src/pages/home/index.jsx
```diff
@@ -1,41 +1,140 @@
...
+          <Typography id="modal-modal-title" variant="h6" component="h2" fontWeight={700}>
+            Your contribution matters
+          </Typography>
+          <Typography id="modal-modal-description" sx={{ mt: 2 }}>
+            Hunger Games is an effort to leverage the results of our OCR. When
+            you use Hunger Games by Open Food Facts, you improve our entire
+            database which helps the community make better food choices. We
+            gathers your annotations and use them to make Open Food Facts more
+            accurate and effecient. <br />
+            <br />
```
nobeeakon, line 131: better use `<p>` to divide text blocks as those are explicit on the start and end

src/pages/home/index.jsx
```diff
@@ -1,41 +1,140 @@
...
+          }}
+        >
+          Learn why your contribution matters
+        </Button>
+      </Box>
+      <Modal
+        open={open}
+        onClose={handleClose}
+        aria-labelledby="modal-modal-title"
+        aria-describedby="modal-modal-description"
```
nobeeakon, line 108: this is a nice usage of aria, however it is better to avoid generic ids because the chance of repeating them is higher and is not very clear to what they are referring to. May be rename to something  like 'modal-contribution-title' and 'modal-contribution-information'

nobeeakon (changes requested):
nice, I'd say that assets could have a better name instead of home1.png and so on :)

sumana2001:
> nice, I'd say that assets could have a better name instead of home1.png and so on :)

Could you suggest something?

nobeeakon:
yeah, sure, if those images are only for the home page, even something a bit more descriptive like home_question.png, home_logos.png, home_ecoscore.png  would work ... what I think is that is better to have a more explicit association between the image and its usage that way if at some point we add more images instead of having a number you would have an idea of what the image is used for



> > nice, I'd say that assets could have a better name instead of home1.png and so on :)
> 
> Could you suggest something?

alexfauquette (approved):
Congrats, that's a much better code than the previous version

@nobeeakon gave you some nice tips about accessibility. 

I doubt on the necessity for an app such as hunger games to be compatible with screen readers. The purpose of the app is to annotate images 🙈

But those are good advice for your other projects (same on naming conventions)

PR #115, opened 2022-08-01
feat: Welcome tour

sumana2001:


src/components/welcome/Buttons.jsx
```diff
@@ -0,0 +1,41 @@
+import React from "react";
+import { Button } from "@mui/material";
+import styled from "styled-components";
```
nobeeakon, line 3: hmmmmm, yeah, this styled components are real nice, yet I guess that for sake of consistency we should stick with Material UI. Yet, if needed you could customize using Material UI features, please have a look at [styled](https://mui.com/material-ui/customization/how-to-customize/), which allow you to modify the styling whitout the expense of using another library

package.json
```diff
@@ -19,6 +19,8 @@
     "react-medium-image-zoom": "^4.4.3",
     "react-router-dom": "^6.3.0",
     "react-scripts": "5.0.1",
+    "reactour": "^1.18.7",
+    "styled-components": "^4.0.0",
```
nobeeakon, line 23: please see the comment below (in Buttons.jsx)

src/components/welcome/Welcome.jsx
```diff
@@ -0,0 +1,132 @@
+import { Button, Box, Typography } from "@mui/material";
+import React, { useState } from "react";
+import Tour from "reactour";
+import Buttons from "./Buttons";
+import QuestionMarkIcon from "@mui/icons-material/QuestionMark";
+
+const Welcome = () => {
+  const steps = [
```
nobeeakon, line 8: think the `steps` could be defined outside of the component, that way we keep the component with only the definitions it needs.

src/components/welcome/Welcome.jsx
```diff
@@ -0,0 +1,132 @@
...
+            Make sure to login to the Open Food Facts website to get credit for
+            your contributions. If you make a mistake, please take the time to
+            correct it by heading to the website.
+          </Typography>
+          <Typography sx={{ mt: 2 }} component="p">
+            In case of any doubt about the games, click on the question mark
+            icon at the top right of the page, and you can reach us on slack at
+            the #hunger-games channel
+          </Typography>
+          <hr />
```
nobeeakon, line 53: if need some additional space I think is better to define it using css, for example wrapping the `Buttons` in a `Box` and adding some padding or margin on top

src/components/welcome/Buttons.jsx
```diff
@@ -0,0 +1,41 @@
...
+    color: #4c5c6b !important;
+    border: 1px solid #4c5c6b;
+  }
+`;
+
+const Buttons = () => {
+  return (
+    <div style={{ display: "flex", justifyContent: "flex-end" }}>
+      <CustomButton
+      // onClick={() => setIsTourOpen(false)}
```
nobeeakon, line 23: What about the onClicks?

src/components/welcome/Buttons.jsx
```diff
@@ -0,0 +1,41 @@
...
+  padding: 0.3em 0.7em !important;
+  fontsize: inherit !important;
+  cursor: pointer !important;
+  margin: 0.2em !important;
+  &:hover {
+    background: white !important;
+    color: #4c5c6b !important;
+    border: 1px solid #4c5c6b;
+  }
+`;
```
alexfauquette, line 17: Maybe just better to do not customize the buttons

src/components/welcome/Buttons.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+const Buttons = () => {
+  return (
+    <div style={{ display: "flex", justifyContent: "flex-end" }}>
+      <CustomButton
+      // onClick={() => setIsTourOpen(false)}
+      >
+        Skip Tour
+      </CustomButton>
+      <CustomButton
```
alexfauquette, line 27: It feels a bit strange to have a log in button in a welcome tour

src/components/welcome/Welcome.jsx
alexfauquette: Yes, as discussed on Monday, the tour could change depending on the current page

alexfauquette (commented):
You already have some feedbacks about the code by itself.

There is two missing part before merging:
- the tour open each time we load the page. Could save in settings for which page the tour has already been done, to not reopen it next time
- after that I propose the following tour:

- question page: 
	- highlight the question: 
		- Answer this question based on the information available on the image.
		- To do so, click on yes/no/skip buttons, or use keys o/n/k.
		- Your answeres will be applied to openfoodfact if you are logged in
	- highlight the filter:
		- You can filter questions from this interface
	- highlight the star:
		- Click on the star to put the current filter settings in favorites.
		- Those favorites will be available in your homepage
	- highlight the remaining questions:
		- Your answers will appear it. In case of error you can go to the openfoodfact website to correct the annotation
- eco-score page:
	- No highlight
		- Those cards are linked to label use in the computation of the eco score.
		- Click on them to answer the remaining question
- All the remaining pages
	- highlight help button:
		- Get some help about the page your are visiting by clicking the help button
	- highlight settings:
		- Modify you language and other settings
	- highlight the links to game in the app bar:
		-  You can access the different games to annotate products by answering questions, or grouping logos

nobeeakon:
I know that this may be out of time, but I feel that in this case (as the UI is not full of options and so on), it may be good to just improve a bit the UI so that things are more intuitive and thus remove the need of a tour. On the other side , think it may be a good alternative to have that help button, and maybe open a modal or something that briefly explains what that page is about. 


 

> You already have some feedbacks about the code by itself.
> 
> There is two missing part before merging:
> 
>     * the tour open each time we load the page. Could save in settings for which page the tour has already been done, to not reopen it next time
> 
>     * after that I propose the following tour:
> 
> 
> ### Proposal for the tour
> 
>     * question page:
>       
>       * highlight the question:
>         
>         * Answer this question based on the information available on the image.
>         * To do so, click on yes/no/skip buttons, or use keys o/n/k.
>         * Your answeres will be applied to openfoodfact if you are logged in
>       * highlight the filter:
>         
>         * You can filter questions from this interface
>       * highlight the star:
>         
>         * Click on the star to put the current filter settings in favorites.
>         * Those favorites will be available in your homepage
>       * highlight the remaining questions:
>         
>         * Your answers will appear it. In case of error you can go to the openfoodfact website to correct the annotation
> 
>     * eco-score page:
>       
>       * No highlight
>         
>         * Those cards are linked to label use in the computation of the eco score.
>         * Click on them to answer the remaining question
> 
>     * All the remaining pages
>       
>       * highlight help button:
>         
>         * Get some help about the page your are visiting by clicking the help button
>       * highlight settings:
>         
>         * Modify you language and other settings
>       * highlight the links to game in the app bar:
>         
>         * You can access the different games to annotate products by answering questions, or grouping logos

alexfauquette:
The problem of a modal is that text is over the interface. So it describes an interface you've never seen. The idea of the tour is to focus attention on the large block and explain why they are here.

In the past we used modal with images, but ofcourse with evolution of the interface, image was outdated, and the description was too long.

But yes, the first step can be an introduction of one sentence about what is the purpose of the game

@teolemon if you have feedbacks

teolemon:
Collapsible inline for the win 
Example of what we do in Smoothie

nobeeakon:
yeah, I mean, maybe just explain what is the purpose of the page/game, and if we manage to make the ui a bit more intuitive I think most of the stuff would be intuitive (which is desirable anyways) and then the need of a tour would be reduced. On the other side the tour should have to be maintained, updated, etc. and introduces other issues like making it consistent across pages, hidding it according to certain decisions, and so on.

alexfauquette:
Ok, so maybe just handle the open on the first visit, and use a simple modal without fancy effects

PR #124, opened 2022-08-05
experiment with login

alexfauquette:
### What
Allows to dev to know if user is logged in or not

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -2,6 +2,7 @@ import React from "react";
 import { createTheme, ThemeProvider } from "@mui/material/styles";
 import CssBaseline from "@mui/material/CssBaseline";
 import { BrowserRouter as Router, Routes, Route } from "react-router-dom";
+import axios from "axios";
 import {
   EcoScorePage,
   LogoAnnotationPage,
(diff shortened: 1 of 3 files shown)


src/App.js
```diff
@@ -46,27 +49,49 @@ const theme = createTheme({
 
 export default function App() {
   const [devMode, setDevMode] = React.useState(getIsDevMode);
+  const [userState, setUserState] = React.useState({
+    userName: "",
+    isLoggedIn: false,
+  });
+
+  const refresh = () => {
```
nobeeakon, line 57: better to have this in the useEffect, that way is simpler to keep track of the dependencies

src/App.js
```diff
@@ -46,27 +49,49 @@ const theme = createTheme({
...
+      .then((rep) => {
+        const cookieUserName = off.getUsername();
+        setUserState({ userName: cookieUserName, isLoggedIn: true });
+      })
+      .catch((err) => {
+        setUserState({ userName: "", isLoggedIn: false });
+      });
+  };
+
+  React.useEffect(refresh, []);
```
nobeeakon, line 71: just wondering what would be a good way of keeping track of the user loggin status... if the effect runs only on mount it will run once, if the effect runs every render if may be too much...

src/App.js
alexfauquette: Thanks for the feedback. The goal is to allow checking the login state at some key point:
- When the user arrives on the page
- When the user answered its 3 first questions (to ask him to log in if not yet)

After we can think about a cleaver way to check cookies, like when not logged in every 3s check is the cookies changed. But that's not the main problem for now.

I'm merging because we do not have the possibility to check that it's working except in production due to some CORS issues

PR #93, opened 2022-07-31
feat: improve Nutri-Score game

alexfauquette:
Fix #63

- The first commit is about
	- Making the game accessible in dev mode ("nutriscore" button in the menu)
	- stick the validate/reject buttons at the  bottom of the page
	- allows modifying image size (out of the issue)
- The second commit is just the result of `yarn prettier`

image

PR #133, opened 2022-08-08
feat: feature flags added to dev mode

### LINKED ISSUE #131: Add feature flags within the dev mode
### What

- [ ] We currently have a dev mode
- [ ] It's currently enabling/disabling all features at once.
- [ ] When enabled, it should show all feature flags, with for each one a enable/disable toggle
- [ ] We could start with:
  - [ ] "Show database"
  - [ ] "Show Nutri-Score game"

sumana2001:
https://user-images.githubusercontent.com/63084088/183458111-d221145f-1a10-4a40-80d5-65bb32020680.mov

Closes #131

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -18,9 +18,7 @@ import {
 import ResponsiveAppBar from "./components/ResponsiveAppBar";
 import DevModeContext from "./contexts/devMode";
 import {
-  getIsDevMode,
-  getShowDatabase,
-  getShowNutriscore,
+  getIsDevMode, getVisiblePages,
 } from "./localeStorageManager";
 import LoginContext from "./contexts/login";
 import off from "./off";
@@ -53,8 +51,7 @@ const theme = createTheme({
 
 export default function App() {
   const [devMode, setDevMode] = React.useState(getIsDevMode);
-  const [showDatabase, setShowDatabase] = React.useState(getShowDatabase);
-  const [showNutriscore, setShowNutriscore] = React.useState(getShowNutriscore);
+  const visiblePages = getVisiblePages();
   const [userState, setUserState] = React.useState({
     userName: "",
     isLoggedIn: false,
@@ -112,14 +109,13 @@ export default function App() {
     <Router>
       <ThemeProvider theme={theme}>
         <LoginContext.Provider value={{ ...userState, refresh }}>
-          <DevModeContext.Provider value={{ 
-            devMode,
-            setDevMode,
-            showDatabase,
-            setShowDatabase,
-            showNutriscore,
-            setShowNutriscore, 
-           }}>
+          <DevModeContext.Provider
+            value={{
+              devMode,
+              setDevMode,
+              visiblePages
+            }}
+          >
             <CssBaseline />
             <ResponsiveAppBar />
             <Routes>
--- a/src/App.js
+++ b/src/App.js
@@ -17,9 +17,7 @@ import {
 } from "./pages";
 import ResponsiveAppBar from "./components/ResponsiveAppBar";
 import DevModeContext from "./contexts/devMode";
-import {
-  getIsDevMode, getVisiblePages,
-} from "./localeStorageManager";
+import { getIsDevMode, getVisiblePages } from "./localeStorageManager";
 import LoginContext from "./contexts/login";
 import off from "./off";
 
@@ -51,7 +49,7 @@ const theme = createTheme({
 
 export default function App() {
   const [devMode, setDevMode] = React.useState(getIsDevMode);
-  const visiblePages = getVisiblePages();
+  const [visiblePages, setVisiblePages] = React.useState(getVisiblePages);
   const [userState, setUserState] = React.useState({
     userName: "",
     isLoggedIn: false,
@@ -113,7 +111,8 @@ export default function App() {
             value={{
               devMode,
               setDevMode,
-              visiblePages
+              visiblePages,
+              setVisiblePages,
             }}
           >
             <CssBaseline />
--- a/src/App.js
+++ b/src/App.js
@@ -16,7 +16,11 @@ import {
 } from "./pages";
 import ResponsiveAppBar from "./components/ResponsiveAppBar";
 import DevModeContext from "./contexts/devMode";
-import { getIsDevMode } from "./localeStorageManager";
+import {
+  getIsDevMode,
+  getShowDatabase,
+  getShowNutriscore,
+} from "./localeStorageManager";
 
 const theme = createTheme({
   palette: {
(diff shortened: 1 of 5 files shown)


src/components/ResponsiveAppBar.jsx
```diff
@@ -43,11 +44,12 @@ const ResponsiveAppBar = () => {
...
-  const { devMode: isDevMode } = React.useContext(DevModeContext);
+  const { showDatabase } = React.useContext(DevModeContext);
+  const { showNutriscore } = React.useContext(DevModeContext);
 
-  const displayedPages = pages.filter(
-    (page) => page.url !== "insights" || isDevMode
-  );
+  const displayedPages = pages
+    .filter((page) => page.url !== "insights" || showDatabase)
+    .filter((page) => page.url !== "nutriscore" || showNutriscore);
```
alexfauquette, line 52: I thought I merge [my PR](https://github.com/openfoodfacts/hunger-games/pull/93/files), but I did not

In the PR I propose to introduce a new parameter `devModeOnly` to the page's objects. Such that we could do 
```jsx
const displayedPages = pages.filter((page) => !page.devModeOnly || isDevMode);
```

Instead of adding a variable for each page, we create, we could use an object `visiblePages` such that `visiblePages = { nutriscore: true, insights: false}` means nutriscore page is visible and insight is not

Then your code would look like

```suggestion
  const { devMode: isDevMode, visiblePages } = React.useContext(DevModeContext);

  const displayedPages = pages
    .filter((page) => !page.devModeOnly || (isDevMode && visiblePages[page.url] ))
```

src/pages/settings/index.jsx
```diff
@@ -60,6 +82,24 @@ export default function Settings() {
...
+            label="Show database"
+            labelPlacement="end"
+          />
+          <FormControlLabel
+            checked={showNutriscore}
+            onChange={handleNutriscoreChange}
+            control={<Switch />}
+            label="Show Nutri-Score game"
+            labelPlacement="end"
+          />
```
alexfauquette, line 100: If you go to object representation, of the visible pages, this part of the code would probably look like that

```suggestion
        {['nutriscore', 'insights'].map(pageUrl => 
        	<FormControlLabel
        		key={pageUrl}
	              checked={visiblePages[pageUrl]}
	              onChange={handleVisiblePagesChange(pageUrl)}
	              control={<Switch />}
	              label={`Show ${pageUrl} page`}
	              labelPlacement="end"
          />)}
```

with `handleVisiblePagesChange = (pageUrl) => (event) => { do the update of pageUrl key depending on the event.target.checked}`

alexfauquette (approved):
I though about another way to manage page visibility, not sure it's better. As you prefer. We can merge without the  changes

src/components/ResponsiveAppBar.jsx
sumana2001: So should I include this new devmode code here or are you merging your other pr?

alexfauquette: I will merge my PR after this one to avoid you git conflicts to manage 😉

src/App.js
```diff
@@ -51,7 +49,7 @@ const theme = createTheme({
 
 export default function App() {
   const [devMode, setDevMode] = React.useState(getIsDevMode);
-  const visiblePages = getVisiblePages();
+  const [visiblePages, setVisiblePages] = React.useState(getVisiblePages);
```
alexfauquette, line 52: You need a setter, to be able to update the value. othersiwe you will not ba able to modify this value

src/localeStorageManager.js
```diff
@@ -39,7 +39,13 @@ export const getIsDevMode = () => {
 
 export const getVisiblePages = () => {
   const settings = localSettings.fetch();
-  return settings[localSettingsKeys.visiblePages] ?? { nutriscore: true, insights: true};
+  console.log({ settings });
```
alexfauquette, line 42:
```suggestion
```

alexfauquette: my bad

src/pages/settings/index.jsx
```diff
@@ -18,21 +18,21 @@ export default function Settings() {
...
+  const { devMode, setDevMode, visiblePages, setVisiblePages } =
     React.useContext(DevModeContext);
 
   const handleDevModeChange = (event) => {
     localSettings.update(localSettingsKeys.isDevMode, event.target.checked);
     setDevMode(event.target.checked);
-    Object.keys(visiblePages).forEach((pageUrl)=>{
-      visiblePages[pageUrl]=event.target.checked
-      localSettings.update(localSettingsKeys.visiblePages[pageUrl], event.target.checked);
-    })
```
alexfauquette: Don't care about that, the pages will just appear as they was before. Not a problem

src/pages/settings/index.jsx
```diff
@@ -18,21 +18,21 @@ export default function Settings() {
...
 
-  const handleVisiblePagesChange = (event, pageUrl) => {
-    localSettings.update(localSettingsKeys.pageUrl, event.target.checked);
-    visiblePages[pageUrl] = event.target.checked;
+  const handleVisiblePagesChange = (pageUrl) => (event) => {
+    const newVisiblePages = {
+      ...visiblePages,
+      [pageUrl]: event.target.checked,
+    };
+    localSettings.update(localSettingsKeys.visiblePages, newVisiblePages);
```
alexfauquette, line 34: The update needs a key/value fort he setting object.

`pageUrl` is not a key of settings, it's a key of `visiblePages`

src/pages/settings/index.jsx
```diff
@@ -18,21 +18,21 @@ export default function Settings() {
...
-    Object.keys(visiblePages).forEach((pageUrl)=>{
-      visiblePages[pageUrl]=event.target.checked
-      localSettings.update(localSettingsKeys.visiblePages[pageUrl], event.target.checked);
-    })
   };
 
-  const handleVisiblePagesChange = (event, pageUrl) => {
-    localSettings.update(localSettingsKeys.pageUrl, event.target.checked);
-    visiblePages[pageUrl] = event.target.checked;
+  const handleVisiblePagesChange = (pageUrl) => (event) => {
```
alexfauquette, line 29: You called `handleVisiblePagesChange(pageUrl)` so with your definition, `handleVisiblePagesChange('nutrition')` leads to 

- `event` has the value `"nutrition"`
- `pageUrl` is `undefined`

That's why you get `undefined:true` in your settings

sumana2001: Oh understood

src/pages/settings/index.jsx
sumana2001: I added this so that when devmode is switched on, regardless of the previous states, both nutrition and insights will be turned on

PR #142, opened 2022-08-11
Send annotations only 5s after the user answer

alexfauquette:
it adds a button in the annotation history that prevent the transmission of the annotation to robotoff, and put back the question on top of the buffer

image

@nobeeakon You will be happy, I started the migration to TS :p

@sumana2001 if you want to see how I didi it, I isolated the feature from typescript in the 3rd commit


src/pages/questions/useQuestionBuffer.ts
```diff
@@ -87,21 +143,58 @@ function reducer(state, action) {
...
         ],
       };
 
+    case "sendAnswers":
+      const minDate = new Date().getTime();
+
+      const newAnswers = state.answers.map((answer) => {
+        const { sendingTime, isPending, validationValue, insight_id } = answer;
+        if (isPending && sendingTime <= minDate) {
+          if (validationValue !== -1) {
```
nobeeakon, line 166: is this SKIPPED_INSIGHT ?

alexfauquette: Yes some of them remain. I will fix it

src/pages/questions/useQuestionBuffer.ts
```diff
@@ -87,21 +143,58 @@ function reducer(state, action) {
...
           ...state.skippedIds,
-          ...(action.payload.value === -1 ? [action.payload.insight_id] : []),
+          ...(action.payload.value === -1 ? [action.payload.insightId] : []),
         ],
       };
 
+    case "sendAnswers":
+      const minDate = new Date().getTime();
+
+      const newAnswers = state.answers.map((answer) => {
```
nobeeakon, line 163: what about filtering (state.answers.filter(..)) the ones that need to be annotated (isPending && sendingTime <= minDate), and use a Promise.all(filteredArray => robotoff.annotate()).then(()=> return {}).catch(()=> return state), something like that...  calling the annotate inside the map makes me feel like doing two different things at the same time

src/pages/questions/useQuestionBuffer.ts
```diff
@@ -87,21 +143,58 @@ function reducer(state, action) {
...
+        }
+        return answer;
+      });
+      return { ...state, answers: newAnswers };
+
+    case "revertAnswer":
+      const answeredQuestionIndex = state.answers.findIndex(
+        ({ insight_id }) => insight_id === action.payload.insightId
+      );
+      const answer = state.answers[answeredQuestionIndex];
```
nobeeakon, line 179: not sure if this could happen, so maybe it is not relevant,... could it happen that it doesn't find it and answeredQuestionIndex === -1? (so that answer === undefined), maybe it is more evident if using answers.find( ... ) instead of answers.findIndex() as the former returns an item or undefined

alexfauquette: It should not occurs, because of the current UI. Only yes/no questions are displayed and so can be reverted

But better to have an early return if no answer correspond to the insight id

src/pages/questions/useQuestionBuffer.ts
nobeeakon, line 163: what about filtering the ones that need to be annotated, and then doing a `Promise.all( filtered( item => robotoff.annotate())).then(res => return { ... }).catch(()=>{})` that way all the annotations are grouped,  I feel that doing the `robotoff.annotate` inside the map is doing two different things at the same time

alexfauquette: I prefer doing it in the map, because like that I send the request and set `isPending` to `false` at the same time.

For now, their is no verification if the request returned an error or not. So I have no await in this code

PR #147, opened 2022-08-14
feat: reporting images added

### LINKED ISSUE #22: [UX] Allows to report images
## Technical aspect

- The interface to mark an image as reported
- A way to manage reported images

### Interface 

To avoid having too much information on the main interface, the buttons to report an image could be added in the image gallery only.
For example a flag `IconButton` in the right top corner of each image

image

### Managing reported images

To discuss with @teolemon :
- Where data is stored
- How do we send it
- How can we access it
- How administrators clean those problematic images

If needed I have a workaround to test the feature: 

On my personal server (amathjourney.com/api), I have an API do save flagged images:

- PUT `https://amathjourney.com/api/off-annotation/flag-image/[code]` with body containing `{imgid, url}`. The `imgid` must contain a number (the id of the image) and `url` is the image src
- GET `https://amathjourney.com/api/off-annotation/flag-image/[code]` will return all the data you put for a given `code`
- DELETE `https://amathjourney.com/api/off-annotation/flag-image/[code]` with body containing `{imgid}`

Basically, it saves all that in a database whit the primary key `code + imgid`

sumana2001:
https://user-images.githubusercontent.com/63084088/184530483-b4beadb6-e1d3-42d3-a94e-83b6a87149ba.mov

Closes #22

### DIFF
--- a/src/pages/questions/ProductInformation.jsx
+++ b/src/pages/questions/ProductInformation.jsx
@@ -13,13 +13,14 @@ import Table from "@mui/material/Table";
 import TableBody from "@mui/material/TableBody";
 import TableCell from "@mui/material/TableCell";
 import TableRow from "@mui/material/TableRow";
-
+import OutlinedFlagIcon from "@mui/icons-material/OutlinedFlag";
+import FlagIcon from "@mui/icons-material/Flag";
 import Zoom from "react-medium-image-zoom";
 import "react-medium-image-zoom/dist/styles.css";
-
 import { useTranslation } from "react-i18next";
 import { NO_QUESTION_LEFT } from "../../const";
 import offService from "../../off";
+import { updateFlagImage, removeFlagImage } from "../../utils";
 
 import {
   localSettings,
@@ -42,6 +43,21 @@ const ProductInformation = ({ question }) => {
   const { t } = useTranslation();
   const [productData, setProductData] = React.useState({});
   const [hideImages, setHideImages] = React.useState(getHideImages);
+  const [flagged, setFlagged] = React.useState([]);
+
+  const flagImage = (src, barcode, index) => {
+    const response = updateFlagImage(barcode, src);
+    const updatedImageArray = [...flagged];
+    updatedImageArray[index] = true;
+    setFlagged(updatedImageArray);
+  };
+
+  const deleteFlagImage = (src, barcode, index) => {
+    const response = removeFlagImage(barcode, src);
+    const updatedImageArray = [...flagged];
+    updatedImageArray[index] = false;
+    setFlagged(updatedImageArray);
+  };
 
   React.useEffect(() => {
     if (!question?.barcode) {
@@ -121,7 +137,11 @@ const ProductInformation = ({ question }) => {
       {!hideImages && productData?.images && (
         <Grid container rowSpacing={1.5} spacing={1}>
           {getImagesUrls(productData.images, question.barcode).map((src) => (
-            <Grid item key={src}>
+            <Grid
+              item
+              key={src}
+              style={{ display: "inline-flex", alignItems: "flex-start" }}
+            >
               <Zoom>
                 <img
                   src={src}
@@ -130,6 +150,19 @@ const ProductInformation = ({ question }) => {
                   style={{ maxWidth: 300, maxHeight: 300 }}
                 />
               </Zoom>
+              {flagged[src[src.length - 5]] ? (
+                <FlagIcon
+                  onClick={() =>
+                    deleteFlagImage(src, question.barcode, src[src.length - 5])
+                  }
+                />
+              ) : (
+                <OutlinedFlagIcon
+                  onClick={() =>
+                    flagImage(src, question.barcode, src[src.length - 5])
+                  }
+                />
+              )}
             </Grid>
           ))}
         </Grid>
--- a/src/pages/questions/ProductInformation.jsx
+++ b/src/pages/questions/ProductInformation.jsx
@@ -29,6 +29,13 @@ import {
   getHideImages,
 } from "../../localeStorageManager";
 
+// src looks like: "https://static.openfoodfacts.org/images/products/004/900/053/2258/1.jpg"
+const getImageId = (src) => {
+  const file = src.split("/").at(-1);
+  const imageId = file.replace(/\..+$/, "");
+  return Number(imageId);
+};
+
 const getImagesUrls = (images, barcode) => {
   if (!images || !barcode) {
     return [];
@@ -40,14 +47,6 @@ const getImagesUrls = (images, barcode) => {
     .map((key) => `${rootImageUrl}/${key}.jpg`);
 };
 
-// src looks like: "https://static.openfoodfacts.org/images/products/004/900/053/2258/1.jpg"
-const getImageId = (src) => {
-  var sliced = src.split("/").at(-1);
-  sliced.replace(/\..+$/, "");
-  const imgid = sliced.slice(0, -4);
-  return Number(imgid);
-};
-
 const ProductInformation = ({ question }) => {
   const { t } = useTranslation();
   const [productData, setProductData] = React.useState({});
@@ -56,27 +55,27 @@ const ProductInformation = ({ question }) => {
 
   const flagImage = (src, barcode) => {
     const imgid = getImageId(src);
-    try {
-      const response = axios.put(
+    axios
+      .put(
         `https://amathjourney.com/api/off-annotation/flag-image/${barcode}`,
         {
           mode: "no-cors",
           imgid: imgid,
           url: src,
         }
-      );
-      const updatedImageArray = [...flagged];
-      updatedImageArray[imgid] = true;
-      setFlagged(updatedImageArray);
-    } catch (err) {
-      console.log(err);
-    }
+      )
+      .catch(() => {
+        console.log("Image flagged");
+      });
+    const newFlagged = [...flagged];
+    newFlagged.push(imgid);
+    setFlagged(newFlagged);
   };
 
   const deleteFlagImage = (src, barcode) => {
     const imgid = getImageId(src);
-    try {
-      const response = axios.delete(
+    axios
+      .delete(
         `https://amathjourney.com/api/off-annotation/flag-image/${barcode}`,
         {
           mode: "no-cors",
@@ -84,19 +83,21 @@ const ProductInformation = ({ question }) => {
             imgid: imgid,
           },
         }
-      );
-      const updatedImageArray = [...flagged];
-      updatedImageArray[imgid] = false;
-      setFlagged(updatedImageArray);
-    } catch (err) {
-      console.log(err);
-    }
+      )
+      .catch(() => {
+        console.log("Something went wrong. Image could not be flagged");
+      });
+    const newFlagged = flagged.filter(
+      (flaggedImageId) => flaggedImageId !== imgid
+    );
+    setFlagged(newFlagged);
   };
 
   React.useEffect(() => {
     if (!question?.barcode) {
       return;
     }
+    setFlagged([]);
     let isStillValid = true;
     setProductData({
       code: question.barcode,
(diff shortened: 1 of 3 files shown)


src/pages/questions/ProductInformation.jsx
```diff
@@ -130,6 +150,19 @@ const ProductInformation = ({ question }) => {
                   style={{ maxWidth: 300, maxHeight: 300 }}
                 />
               </Zoom>
+              {flagged[src[src.length - 5]] ? (
+                <FlagIcon
+                  onClick={() =>
+                    deleteFlagImage(src, question.barcode, src[src.length - 5])
```
nobeeakon, line 156: what about passing the `src` and `question.barcode` to `deleteFlagImage` , and in `deleteFlagImage` obtain the `index` (imageId?), and that same `index` can be passed down to `updateFlagImage()` which seems to be needing it.

src/utils.js
```diff
@@ -41,3 +43,26 @@ export const removeEmptyKeys = (obj) => {
 export function sleep(ms) {
   return new Promise((resolve) => setTimeout(resolve, ms));
 }
+
+export const updateFlagImage = (barcode, src) => {
+  return axios.put(
+    `https://amathjourney.com/api/off-annotation/flag-image/${barcode}`,
+    {
+      mode: "no-cors",
+      imgid: Number(src[src.length - 5]),
```
nobeeakon, line 52: would be useful to have somewhere a comment showing how the `src` looks like, to have an idea of why/how this code (`src[src.length - 5]`) works... also, is there any case where the number consist in more than 1 digit?

src/utils.js
```diff
@@ -41,3 +43,26 @@ export const removeEmptyKeys = (obj) => {
 export function sleep(ms) {
   return new Promise((resolve) => setTimeout(resolve, ms));
 }
+
+export const updateFlagImage = (barcode, src) => {
```
nobeeakon, line 47: just wondering here (no clear idea), but are these two functions going to be used `updateFlagImage` and `removeFlagImage` in another place besides the questions? mmm think is better to have things as close to the place they are used and feel that having it here in a generic utils gets things a bit out of context. Please ignore this comment if you feel it is not very relevant

src/utils.js
```diff
@@ -41,3 +43,26 @@ export const removeEmptyKeys = (obj) => {
...
+    `https://amathjourney.com/api/off-annotation/flag-image/${barcode}`,
+    {
+      mode: "no-cors",
+      imgid: Number(src[src.length - 5]),
+      url: src,
+    }
+  );
+};
+
+export const removeFlagImage = (barcode, src) => {
```
nobeeakon, line 58: maybe would be useful to have a comment of what means removing or updating a flag, just to give some context of the _why_ of this operation

src/pages/questions/ProductInformation.jsx
```diff
@@ -42,6 +43,21 @@ const ProductInformation = ({ question }) => {
   const { t } = useTranslation();
   const [productData, setProductData] = React.useState({});
   const [hideImages, setHideImages] = React.useState(getHideImages);
+  const [flagged, setFlagged] = React.useState([]);
+
+  const flagImage = (src, barcode, index) => {
+    const response = updateFlagImage(barcode, src);
```
nobeeakon, line 49: considering that the `updateFlagImage` is doing an asyc call, you may want to do the `flagImage` an `async` ( ...`flagImage = async (...) => { ... }` ) function, and `await` the response of the `updateFlagImage`

nobeeakon (commented):
really good @sumana2001 ... maybe it would be nice to have a tooltip in the button or a question mark icon with a tooltip next to the flag or something that explains what the flag means

src/pages/questions/ProductInformation.jsx
```diff
@@ -130,6 +150,19 @@ const ProductInformation = ({ question }) => {
...
               </Zoom>
+              {flagged[src[src.length - 5]] ? (
+                <FlagIcon
+                  onClick={() =>
+                    deleteFlagImage(src, question.barcode, src[src.length - 5])
+                  }
+                />
+              ) : (
+                <OutlinedFlagIcon
+                  onClick={() =>
```
nobeeakon, line 161: better to use an [IconButton](https://mui.com/material-ui/api/icon-button/#main-content) and associate the `onClick` to the button

src/pages/questions/ProductInformation.jsx
alexfauquette: the helper `getImagesUrls` only return image src, but it can be modified to return, all the necessary information. Especially imageId which will avoid doing the `src[src.length - 5]`

src/utils.js
alexfauquette: For now I don't see where else it could be used.

About the `src[src.length - 5]` I added a comment about how it can be avoid. So it will be possible to pass both `src` and `imgid` to this function

alexfauquette: Yes, of course as soon as a product has more than 9 images, it uses 2 digits numbers 😉

src/pages/questions/ProductInformation.jsx
sumana2001: Is there a better way to get the imgid than `src[src.length - 5]`? Since it is not handling the double digit cases?

src/utils.js
sumana2001: I did that so that the API calls are separate. Should I just include it in the `removeFlagImage` file?

src/pages/questions/ProductInformation.jsx
sumana2001: Since that might take some time, I thought it would be better to work in the background to improve the user experience.

src/pages/questions/ProductInformation.jsx
```diff
@@ -151,17 +153,29 @@ const ProductInformation = ({ question }) => {
                 />
               </Zoom>
               {flagged[src[src.length - 5]] ? (
-                <FlagIcon
-                  onClick={() =>
-                    deleteFlagImage(src, question.barcode, src[src.length - 5])
-                  }
-                />
+                <Tooltip title="Unflag Image">
```
nobeeakon, line 156: Maybe something a bit more informative?, for example I don't have a clear idea of what flagging an image means :(

src/pages/questions/ProductInformation.jsx
```diff
@@ -151,17 +153,29 @@ const ProductInformation = ({ question }) => {
...
+                  <IconButton>
+                    <FlagIcon
+                      onClick={() =>
```
nobeeakon, line 159: nit: please put the `onClick` in the `IconButton`

src/pages/questions/ProductInformation.jsx
nobeeakon: ohh I see, that's a nice idea, something on the lines of [optimistic updates](https://stackoverflow.com/questions/33009657/what-is-optimistic-updates-in-front-end-development), I think as you said it is good UX. An option would be to keep it as you have it, but please add the `updateFlagImage(...).catch()` and in the `catch` you would need to set it back to the original state. Another option would be to wait and show a loading spinner in the button, I think in most cases it shouldn't take too long.

src/pages/questions/ProductInformation.jsx
nobeeakon: what about splitting, taking the last part and then removing the file extension, something like:
`src = "https://static.openfoodfacts.org/images/products/004/900/053/2258/1.jpg"`
`src2= src.split("/").at(-1)`
`src2.replace(/\..+$/, "")` or maybe removing the non digits?

please whenever you use it, please add a comment like
`// src looks like: "https://static.openfoodfacts.org/images/products/004/900/053/2258/1.jpg" `
that way we don't have to remember how the src was

sumana2001:
I have made the changes you suggested, please have a look again. @nobeeakon @alexfauquette

src/pages/questions/ProductInformation.jsx
```diff
@@ -138,47 +170,41 @@ const ProductInformation = ({ question }) => {
...
-                        flagImage(src, question.barcode, src[src.length - 5])
-                      }
-                    />
-                  </IconButton>
-                </Tooltip>
-              )}
-            </Grid>
-          ))}
+          {getImagesUrls(productData.images, question.barcode).map(
+            (src, imgid) => (
```
alexfauquette, line 174: You introduce `imgid` that is in fact the array index (so not sure it matches with the id). My proposal was to modify `getImagesUrls` such that you can write

```suggestion
            ({ src, imgid }) => (
```

src/pages/questions/ProductInformation.jsx
```diff
@@ -38,10 +40,58 @@ const getImagesUrls = (images, barcode) => {
...
+        `https://amathjourney.com/api/off-annotation/flag-image/${barcode}`,
+        {
+          mode: "no-cors",
+          imgid: imgid,
+          url: src,
+        }
+      );
+      const updatedImageArray = [...flagged];
+      updatedImageArray[imgid] = true;
+      setFlagged(updatedImageArray);
```
alexfauquette, line 70: I don't get what you're doing. You have an array of flagged ids. But how is it reset? For example I flag the second image. After answering the question I get a new product which should not be flagged

Maybe a useEffect to reset to an empty array would do the job

src/pages/questions/ProductInformation.jsx
```diff
@@ -120,18 +170,41 @@ const ProductInformation = ({ question }) => {
...
+              >
+                <Zoom>
+                  <img
+                    src={src}
+                    alt=""
+                    loading="lazy"
+                    style={{ maxWidth: 300, maxHeight: 300 }}
+                  />
+                </Zoom>
+                {flagged[src[src.length - 5]] ? (
```
alexfauquette, line 188: If `imgid` becomes reliable
```suggestion
                {flagged[imgid] ? (
```

src/pages/questions/ProductInformation.jsx
```diff
@@ -120,18 +170,41 @@ const ProductInformation = ({ question }) => {
...
+                    <IconButton
+                      onClick={() => deleteFlagImage(src, question.barcode)}
+                    >
+                      <FlagIcon />
+                    </IconButton>
+                  </Tooltip>
+                ) : (
+                  <Tooltip title="Something wrong? Report/flag the image!">
+                    <IconButton
+                      onClick={() => flagImage(src, question.barcode)}
```
alexfauquette, line 199:
```suggestion
                      onClick={() => flagImage(src, question.barcode, imgid)}
```

src/pages/questions/ProductInformation.jsx
```diff
@@ -120,18 +170,41 @@ const ProductInformation = ({ question }) => {
...
+                    src={src}
+                    alt=""
+                    loading="lazy"
+                    style={{ maxWidth: 300, maxHeight: 300 }}
+                  />
+                </Zoom>
+                {flagged[src[src.length - 5]] ? (
+                  <Tooltip title="Unflag the image">
+                    <IconButton
+                      onClick={() => deleteFlagImage(src, question.barcode)}
```
alexfauquette, line 191:
```suggestion
                      onClick={() => deleteFlagImage(src, question.barcode, imgid)}
```

alexfauquette (commented):
I notice that you do not fetch the flagged images for a product to check if some images have already been flagged by other users.

It's not a problem, because it should be rare, and I prefer having rare inconsistencies than spamming my toy server 🙈

src/pages/questions/ProductInformation.jsx
```diff
@@ -38,10 +40,58 @@ const getImagesUrls = (images, barcode) => {
     .map((key) => `${rootImageUrl}/${key}.jpg`);
 };
 
+// src looks like: "https://static.openfoodfacts.org/images/products/004/900/053/2258/1.jpg"
+const getImageId = (src) => {
+  var sliced = src.split("/").at(-1);
+  sliced.replace(/\..+$/, "");
+  const imgid = sliced.slice(0, -4);
+  return Number(imgid);
```
nobeeakon, line 48: nit: please try instead
(note: better avoid using `var`)
`const file = src.split("/").at(-1);`
`const imageId = file.replace(/\..+$/, "");`
`return Number(imageId);`

src/pages/questions/ProductInformation.jsx
```diff
@@ -38,10 +40,58 @@ const getImagesUrls = (images, barcode) => {
...
+    const imgid = getImageId(src);
+    try {
+      const response = axios.put(
+        `https://amathjourney.com/api/off-annotation/flag-image/${barcode}`,
+        {
+          mode: "no-cors",
+          imgid: imgid,
+          url: src,
+        }
+      );
```
nobeeakon, line 67: nit: you could try `axios.put( ... ).catch(()=>{} )` (no need to log the error), also no need to have the `response` variable if not used. [axios error handling docs](https://axios-http.com/docs/handling_errors)

src/pages/questions/ProductInformation.jsx
nobeeakon: mmm, what about not setting to `true` (it is not an `{}` ), and just do `const newFlagged = [...flagged]` and `newFlagged.push(imageId)` (if possible avoid abbreviations, at some point they get confusing to remember). Later when using it you could, instead of doing `flagged[src[src.length - 5]]`, you could do something like `flagged.includes(imageId)`

src/pages/questions/ProductInformation.jsx
nobeeakon: `imgid` is actually not used, if possible please remove it

src/pages/questions/ProductInformation.jsx
nobeeakon: this could be `flagged.includes(getImageId(src))`, please check the comment above

src/pages/questions/ProductInformation.jsx
```diff
@@ -38,10 +40,58 @@ const getImagesUrls = (images, barcode) => {
...
+        `https://amathjourney.com/api/off-annotation/flag-image/${barcode}`,
+        {
+          mode: "no-cors",
+          data: {
+            imgid: imgid,
+          },
+        }
+      );
+      const updatedImageArray = [...flagged];
+      updatedImageArray[imgid] = false;
```
nobeeakon, line 89: based on  the comment above, this could be `const newFlagged = flagged.filter( flaggedImageId => flaggedImageId !== imageId)` and `setFlagged(updatedImageArray)`

src/pages/questions/ProductInformation.jsx
```diff
@@ -42,11 +51,53 @@ const ProductInformation = ({ question }) => {
...
+        `https://amathjourney.com/api/off-annotation/flag-image/${barcode}`,
+        {
+          mode: "no-cors",
+          data: {
+            imgid: imgid,
+          },
+        }
+      )
+      .catch(() => {
+        console.log("Something went wrong. Image could not be flagged");
```
nobeeakon, line 88: nit: no need for the `console.log` as it is not useful for the client, you could leave `.catch(() => {})` and it will be fine

nobeeakon (approved):

PR #154, opened 2022-08-16
feat: Nutrition game - table, tableRow and select Unit components

Shmigolk:
image
I added table component, tableRow component and select Unit component. Currently there is no ability to add Nutriments but I will add it later. 

To imitate POST Request I log to the console an array of nutriments with values, or if all values are empty strings I  log a message to the console "You forgot to input values dide"


src/pages/nutrition/index.jsx
```diff
@@ -18,7 +40,18 @@ export default function Nutrition() {
...
-      <ProductNutriments setNutriments={{}} nutriments={[]} />
+      <ProductNutriments
+        setNutriments={setNutriments}
+        nutriments={nutriments}
+      />
+      <NutritionTable
+        nutriments={nutriments.filter((nutr) => nutr.display)}
+        setNutriments={setNutriments}
+        //additionalNutriments={nutriments.filter((nutr) =>
+        // !nutr.display)}
```
alexfauquette, line 51:
```suggestion
```

src/pages/nutrition/productCard.jsx
```diff
@@ -37,6 +39,14 @@ export default function ProductNutriments({ setNutriments, nutriments }) {
       setIndex(0);
       setPage((prevPage) => ++prevPage);
     }
+    const resArr = nutriments.filter(
+      (nutr) => nutr.display && +nutr.value && +nutr.value > 0
+    );
```
alexfauquette, line 44: What does `resArr` mean?

A value can be 0. For example knowing that some coca has 0 sugar is an information
```suggestion
    const resArr = nutriments.filter(
      (nutr) => nutr.display && nutr.value !== null && nutr.value >= 0
    );
```

src/pages/nutrition/tableRow.jsx
```diff
@@ -0,0 +1,84 @@
...
+      <Box
+        sx={{
+          display: "flex",
+        }}
+      >
+        <SelectUnit
+          options={nutriment.quantification}
+          value={"quantification"}
+          onchangeHandler={onchangeHandler}
+        />
```
alexfauquette, line 42: You use two ties `SelectUnit` and not always to select unit

Why not directly put the `Select` here

```suggestion
        <TextField select value={nutriment.quantification}>
        	<MenuItem value="=">=</MenuItem>
  	</TextField>
```

Shmigolk: The problem was that we have different unit. But now I think it is better to add maybe unitsList property into nutrients data

alexfauquette (approved):

PR #173, opened 2022-08-28
feat: selectable logo annotation and add checkbox

### LINKED ISSUE #139: Make label text selectable
### Story
- Make label text selectable
image

### Part of
- #85

alexfauquette:
Fix #139 and some lines of #85

add a check box with same behavior as clicking on the image

remove the text from the action area: clicking on it do nothing, and so you can copy past labeled cards

image


teolemon (approved):
Nice ideas

PR #239, opened 2022-10-01
Fix insights service call for non-annotated search

### LINKED ISSUE #230: Wrong request when searching Insight without annotation
### What

In insight preview, if I choose non annotated in annotation select, I don't get a result.

This is because it send the request 
https://robotoff.openfoodfacts.org/api/v1/insights?annotation=not_annotated&page=1&annotated=0&count=25
which triggers a 400 error on robotoff side.

It should not send the annotation parameter
https://robotoff.openfoodfacts.org/api/v1/insights?page=1&annotated=0&count=25
works.

Adnan-Sait:
### What
- Fixes Insights service call for non-annotated search. The annotation parameter is not sent when the annotation is "non-annotated".

### Screenshot
hunger-games-fix

### Fixes bug(s)
- Fix #230 

### Part of 
- #230

### DIFF
--- a/src/robotoff.ts
+++ b/src/robotoff.ts
@@ -133,6 +133,7 @@ const robotoff = {
     let annotated;
     if (annotation.length && annotation === "not_annotated") {
       annotated = "0";
+      annotation = "";
     }
     return axios.get(`${ROBOTOFF_API_URL}/insights`, {
       params: removeEmptyKeys({


alexfauquette (approved):

alexfauquette:
Thanks for your first contribution :)

PR #248, opened 2022-10-04
User feedbacks: Allows to keep an eye on the form in logos game

alexfauquette:
The form stays sticky to the top

image

And add a small button at the bottom right to scroll top

image


VaiTon:
My personal idea: maybe adding a shadow when the page is not on the top (after scrolling) could improve the look. Good job though!

teolemon (approved):

PR #257, opened 2022-10-06
chore: upgrade dependencies

alexfauquette:


VaiTon:
I would advise against upgrading every dependency in a single PR. It's better to let dependabot handle it and merge PRs as they come IMHO

PR #258, opened 2022-10-09
Use translations everywhere

### LINKED ISSUE #246: Use i18n for all texts
Some texts are not translated, mostly because of my fault. Could be good to clean that.

We are using `"react-i18next"`. To know how to use it, the simplest is maybe to search for a file with `const { t } = useTranslation();`

To add a translation key, you should add it to `src/i18n/common.json` and `src/i18n/en.json`

dadolhay:
### What
Changes all hardcoded texts to translations

### Screenshot
No visual changes

### Part of 
closes #246


src/components/welcome/Welcome.jsx
```diff
@@ -84,35 +77,46 @@ const getSteps = (withSelector) => [
...
-              Answer simply with a <strong>yes/no</strong>! Don't know the
-              answer? That's alright, just skip it.
+              {t("helper.welcome.page2.text1")}
             </Typography>
+            <Typography
+              component="p"
+              sx={{ mt: 2 }}
+              dangerouslySetInnerHTML={{
+                // https://github.com/i18next/react-i18next/issues/189#issuecomment-235181562
+                __html: t("helper.welcome.page2.text2"),
```
alexfauquette, line 95: What about using `<Trans/>` https://react.i18next.com/latest/trans-component#using-with-react-components ?

alexfauquette (commented):
Sounds really good :)

Maybe we can avoid the HTML integration. Will have a deeper look tomorrow

src/components/welcome/Welcome.jsx
dadolhay: I did not want to use the Trans element for consistency reasons, but from your words, I take it that you are OK with the change

dadolhay:
I updated the PR, migrated to the "alternate use" of the `Trans` component, no more "dangerous" stuff

alexfauquette (approved):
It works perfectly 👌

PR #293, opened 2022-10-18
Feat: Show product quantity and labels

alexfauquette:
Small code styling clean up

In the product data; if value is empty show a `?` and add the product labels and quantity

image


alexfauquette:
@raphael0202 Not sure if it was the quantity or serving size you were looking for

raphael0202:
it's `quantity`, that's the field we update it `product_weight` questions

PR #297, opened 2022-10-20
Feat: Improve agri-balise categories cards

alexfauquette:
- [x] link to questions filtered by country
- [x] allows to load more categories
- [x] use the categories translations
- [ ] use images from wikidata


alexfauquette:
image

teolemon:
That's great 👌 can you use the country taxo for the country selector?

PR #320, opened 2022-10-24
Feat: Allows to fetch logos from some categories

alexfauquette:
To be done:

- [x] Data fetching
- [x] Annotation POST request annotation form distinct from search form?
- [ ] UX refinement

[preview](https://deploy-preview-320--warm-cannoli-79bbb2.netlify.app/logos/product-search?tag=en%3Anutriscore-grade-a) Notice that i's netlify so you can not be logged in


alexfauquette:
So on top, you have a for to fetch products according to one filter

image


When you press annotate, you will see selected logos and a form to annotate them

image

@raphael0202 @teolemon If you have insights to improve it. Otherwise you can redirect the PR to master and merge

raphael0202:
Amazing 🤩
The UX is perfect! 

A few remarks:
- Is it possible to have label autocomplete on search page?
- the UI doesn't display well on my mobile


I think we should add it in to the menu, this is really valuable

alexfauquette:
For the mobile interface, I let that for later. I honestly do not recommend to use it on mobile

PR #325, opened 2022-10-25
feat: localize keyboard shortcut

benbenben2:
### What
Replaced the shortcut to answer questions in Hunger Games to "y"

### Why
The actual shortcut to reply "yes" to questions in Hunger Games is "o".
Assuming that it is a typo (French abbreviation maybe?), I changed it to "y" (for yes).

**Remark**: it is my first PR to hunger games project. I tested the changes locally.


teolemon:
Ideally, yes it should be localizable, and we should carefully choose patterns. 
I wonder whether libraries exist for such a common issue ( yes no maybe)

alexfauquette:
Don't need a library.

A file `shortcut-l10n.js`

```js
const SHORTCUT_LOCALISATION = {
	en: {
		yes: "y",
		no: "n",
		skip: "k",
	},
	fr: {
		yes: "y",
		no: "n",
		skip: "k",
	}
}

const getShortcuts = (lang) => {
	return {...SHORTCUT_LOCALISATION.en, ...SHORTCUT_LOCALISATION[lang ?? getLang()]}
}
```

That in the main file you can do 

```
const shortcuts = getShortcuts()

// ...

{t("questions.yes")} ({shortcuts.yes})
```


probably need to replace the `switch(event.keyCode)` by `switch(event.code)`

benbenben2:
Hi guys,
Sorry for late reply. 
I will not be able to do that by myself :-( 
Should I close this PR and let you do this by yourself?

alexfauquette:
@benbenben2 No problem, I will finish the PR

VaiTon:
Hey! Nice idea about the localized shortcuts! 

Something I would like to add is the fact that a lot of localized cli commands always take y as yes and n as no, so that if somebody is used to the english version it would work too.

alexfauquette:
@benbenben2 If you want to have a look, I added a commit to modify shortcuts depending on localization

PR #411, opened 2022-12-23
feat: allows to see packaging interface for specific products

### LINKED ISSUE #396: It would be helpful if we could view an arbitrary barcode in this game (as in type it it, or pass it through the URL)
### What
- It would be helpful if we could view an arbitrary barcode in this game (as in type it it, or pass it through the URL)

### Mockups
image

alexfauquette:
Fix #396 by allowing to pass parameter code in the URL

hunger.openfoodfacts.org/packaging?code=3245414079106

Also adding `country` and `creator` parameters

+ some minor fixes

### DIFF
--- a/src/pages/packaging/index.tsx
+++ b/src/pages/packaging/index.tsx
@@ -27,6 +27,7 @@ import ZoomableImage from "../../components/ZoomableImage";
 import { getImagesUrls } from "../questions/utils";
 import offService from "../../off";
 import { useTranslation } from "react-i18next";
+import useUrlParams from "../../hooks/useUrlParams";
 
 type CustomProps = {
   options: Option[];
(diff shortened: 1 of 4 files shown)


teolemon (approved):
That's 💯 @alexfauquette

src/pages/packaging/index.tsx
```diff
@@ -243,11 +244,16 @@ const Page = () => {
   const packagingMaterials = useOptions("packaging_materials", lang);
   const packagingShapes = useOptions("packaging_shapes", lang);
   const packagingRecycling = useOptions("packaging_recycling", lang);
+  const [searchState] = useUrlParams({
+    country: "en:france",
```
VaiTon, line 248: I would default this to world

alexfauquette: The reason is to do as many as possible products for the french agency ADEM and open it to the world after that

It's kind of related to the issue asking to add a field country in the settings such that

PR #412, opened 2022-12-24
ci: use actions/setup-node when deploying

VaiTon:


.github/workflows/deploy.yml
```diff
@@ -1,36 +1,43 @@
...
 
-# Controls when the workflow will run
 on:
   # Triggers the workflow on push or pull request events but only for the master branch
   push:
     branches: [master]
 
+  pull_request:
+    branches: [master]
+
```
alexfauquette, line 10: Are you sure? If I'm correct, it will deploy in production as soon as we open a PR on master

.github/workflows/deploy.yml
```diff
@@ -1,36 +1,43 @@
...
+      - name: Build
+        run: yarn build
+
+      - name: Deploy
+        run: yarn deploy
 
-        # --legacy-peer-deps is used due to a dependency between "@vue/cli-plugin-eslint" and eslint
-      - name: Deploy 🚀
+      - name: Deploy to github pages 🚀
+        if: github.ref == 'refs/heads/master'
```
alexfauquette, line 40: What's the problem solved by the `if: github.ref == 'refs/heads/master'`

VaiTon: The one you mentioned in the other comment. The deployment to github pages is done only when running from the master branch, not from a PR.

.github/workflows/deploy.yml
VaiTon: See my comment below the other observation :)

.github/workflows/deploy.yml
VaiTon: In fact you can see that it has been skipped for this PR

alexfauquette: > you can see that it has been skipped for this PR

Not 100% sure, but I think GitHub run actions according to the main branch. The only way I know to test actions is to push it to the main branch of your forked repository, and open a PR on it.


I still don't get the problem you're solving. For me, you do two things:
- trigger deploy action when opening a PR on `master` by adding
    ```yaml
    pull_request:
      branches: [master]
    ```
- block the deployment by adding 
    ```yaml
    if: github.ref == 'refs/heads/master'
    ```

Notice that no `pull_request` could get this `github.ref`, because for PR ref is `refs/pull/<pr_number>/merge`

> The ref given is fully-formed, meaning that for branches the format is `refs/heads/<branch_name>`, for pull requests it is `refs/pull/<pr_number>/merge`, and for tags it is `refs/tags/<tag_name>`. For example, refs/heads/feature-branch-1.

From: https://docs.github.com/en/actions/learn-github-actions/contexts#github-context


----

Why not removing both 
```yaml
pull_request:
  branches: [master]
```

and

```yaml
if: github.ref == 'refs/heads/master'
```

VaiTon: This is useful to test for changes before merging with the same workflow (except deploying) that we use for a real deployment. I have no hard view about it, so if you want it I can remove it. Also about the test, if you want we can try in my repo. Let me know!

alexfauquette (approved):
Thanks for the explanation, I did not though about running this action just to see when it fails  :)

Issue #422, opened 2022-12-26
Add a "Select All" button in Infinite mode


teolemon:
### Problem
Add a "Select All" button  in Infinite mode
### Part of
- #85

PR #450, opened 2023-01-03
feat: select and deselect all button for infinite mode

diivi:
### What
- Added a button to select/deselect all the entries in infinite mode to annotate them

### Screenshot
image
image

### Fixes bug(s)
#422

### DIFF
--- a/src/pages/logos/LogoDeepSearch.jsx
+++ b/src/pages/logos/LogoDeepSearch.jsx
@@ -44,6 +44,7 @@ export default function LogoSearch() {
 
   const [annotatedLogos, setAnnotatedLogos] = React.useState([]);
   const [logosToAnnotate, setLogosToAnnotate] = React.useState([]);
+  const [selectAll, setSelectAll] = React.useState(false);
   // TODO: allows to fetch more when reaching data limit
   const [searchCount] = React.useState(DEFAULT_COUNT);
   const [searchState, setSearchState] = useUrlParams({ type: "", value: "" });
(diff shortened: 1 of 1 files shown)


src/pages/logos/LogoDeepSearch.jsx
```diff
@@ -88,6 +89,18 @@ export default function LogoSearch() {
     };
   }, [searchState, searchCount]);
 
+  React.useEffect(() => {
+    if (selectAll) {
+      setLogosToAnnotate((prev) =>
+        prev.map((logo) => ({ ...logo, selected: true }))
+      );
```
alexfauquette, line 96: This suffers from one problem: pagination.

This select all will also select logos on the previous page and the next page that you do not see currently. It could work if restricted to the current page logos

src/pages/logos/LogoDeepSearch.jsx
```diff
@@ -90,16 +90,38 @@ export default function LogoSearch() {
...
       setLogosToAnnotate((prev) =>
-        prev.map((logo) => ({ ...logo, selected: false }))
+        // set current page to unselected, keep old
+        prev.map((logo, index) => ({
+          ...logo,
+          selected:
+            index < page * pageSize && index >= (page - 1) * pageSize
+              ? false
+              : logo.selected,
+        }))
```
alexfauquette, line 113: I would keep the previous behavior for the unselect all.
```suggestion
        prev.map((logo) => ({ ...logo, selected: false }))
```

It's harder to correct an error than to add a new annotation. So I prefer that 
- **select all* is limited to what the user can see
- **unselect all** is extended to what users do not see and might have forget

But that's a personal opinion

alexfauquette (commented):
Sounds pretty good, just added a last comment :)

alexfauquette:
I just tried it, and their is few UX that are disturbing.

First, when you load the page, the button allows you to "Unselec all" even if nothing is currently selected. If you do select all and go to the next page, it unselects the previous page.

https://user-images.githubusercontent.com/45398769/210622233-a378b4b5-b5cc-458b-bf54-58fe4e6800bb.mp4


Maybe the management with a state+effect is not the best. What about two buttons with "select all" and "unselect all"?

alexfauquette (approved):
Works supper well 👍🎉

Congrats for your first contribution :)

Issue #456, opened 2023-01-04
The recycling dashboard links to labels vs packaging


teolemon:
### What
The recycling dashboard links to labels vs packaging
https://hunger.openfoodfacts.org/logos/deep-search?type=label&value=en:10-NiCD
https://hunger.openfoodfacts.org/logos/deep-search?type=packaging&value=en:01-PET

alexfauquette:
How to solve:
in `LogoDefinition` as a `type` property which can be either 

```ts
type InsightType = 'label' | 'product_weight' | 'category' | 'expiration_date' | 'packager_code' | 'brand' | 'packaging' | 'qr_code'

export type LogoDefinition = {
  ...
  type: InsightType;
}
```

In `DashboardCard.ts` update `questionsUrl` and `logoAnnotationUrl` to use this new type property

In `LogoQuestionValidator.jsx` this new property will be available in `selectedOption` and should be used in the definition of `filterState`

psankhe28:
I tried solving this issue by taking the help of the above-mentioned solution but after implementing steps 1 and 2, I am encountering the following error:
![image](https://user-images.githubusercontent.com/84843461/210585417-6c65d7fa-3563-4fbf-b18d-233218e6355d.png)

psankhe28:
Okay. Thanks and sorry for the inconvenience.

alexfauquette:
Do you want to open the PR and continue the issue?

psankhe28:
Yeah!

alexfauquette:
Fixed by #457

PR #457, opened 2023-01-05
Fix: Recycling dashboard links to labels vs packaging

psankhe28:
Part of #456 

### What
I tried to implement the steps as mentioned in the comment section of the issue #456 .

### Getting this error

image

### DIFF
--- a/src/pages/logosValidator/dashboardDefinition.ts
+++ b/src/pages/logosValidator/dashboardDefinition.ts
@@ -1,4 +1,13 @@
-type InsightType = 'label' | 'product_weight' | 'category' | 'expiration_date' | 'packager_code' | 'brand' | 'packaging' | 'qr_code'
+type InsightType =
+  | "label"
+  | "product_weight"
+  | "category"
+  | "expiration_date"
+  | "packager_code"
+  | "brand"
+  | "packaging"
+  | "qr_code";
+
 export type LogoDefinition = {
   /**
    * The OpenFoodFacts tag used or interactions with robotoff and off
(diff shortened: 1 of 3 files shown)


alexfauquette:
Thanks for the PR, now it makes sense. You added `type` as a required property (nice 👍)

But now, `DashboardCard` is expected to get `type` in its props. You have two solutions:

- Make `type` optional with `type?: ....` and set a default value (I would not recommend this because there is no default value that is more legitimate than another)
- Add a `type` to every item in `LOGOS`

Are you familiar with shortcuts for fast editing in your IDE?
If not, to avoid spending too much time on this, you can do the following replacement :

Replace `logo: "` by `type: "label", logo: "`

abughalib:
Doing something like this would resolve the issue
```jsx
...
interface Logos {
  [key: string]: LogoDefinition;
}

export const LOGOS: Logos = {
  // Nutriscore
  "en:nutriscore-grade-a": {
    tag: "en:nutriscore-grade-a",
    label: "Nutriscore A",
    logo: "https://static.openfoodfacts.org/images/attributes/nutriscore-a.svg",
    predictor: null,
    type: 'label' // Label
  },
...
```

abughalib:
> @alexfauquette Did the changes by adding type property to LOGOS, but still facing the same error. Can you please check it out once.

Please have a look at my previous message which should fix the issue.
https://github.com/openfoodfacts/hunger-games/pull/457#issuecomment-1372005076

src/pages/logosValidator/dashboardDefinition.ts
```diff
@@ -19,682 +20,707 @@ export type LogoDefinition = {
...
    */
   link?: string;
+  type: InsightType;
   /**
    * To specify if the game needs to be restricted to a specific robotoff predictor
    * @default "universal-logo-detector"
    */
   predictor?: string;
 };
-
```
abughalib, line 28: interface Logos {
  [key: string]: LogoDefinition;
}

src/pages/logosValidator/dashboardDefinition.ts
```diff
@@ -19,682 +20,707 @@ export type LogoDefinition = {
...
 export const LOGOS = {
```
abughalib, line 30:
```suggestion
export const LOGOS: Logos  = {
```

abughalib (changes requested):
Try those suggested changes, and run Prettier Code Formatter if you can or other if specified.

src/pages/logosValidator/dashboardDefinition.ts
psankhe28: Sure. Will try

psankhe28:
It worked. Thanks @abughalib

src/pages/logosValidator/dashboardDefinition.ts
alexfauquette: The is that latter on the keys are used to let Typesctip detect speeling mistake and add autocompletion on VSCode
```
logos: (keyof typeof LOGOS)[];
```

With your solution key type is a string, so it breaks this check. I added a commit with the following solution:

```ts
type LogoType = { [key in keyof typeof UNTYPED_LOGOS]: LogoDefinition };

export const LOGOS = UNTYPED_LOGOS as LogoType;
```

alexfauquette (approved):
Sounds good, I merge, and you can do the 3rd steps in a distinct PR

Issue #458, opened 2023-01-05
Changing styles of buttons in Questions Page.


Apurv428:
As mentioned in the issue 251, the task is as follows:

![image](https://user-images.githubusercontent.com/84929607/210760013-8ab803bb-e55d-4fc3-a905-744919a1edcc.png)

Apurv428:
I would like to work on this issue.

Issue #251, opened 2022-10-05
Yes/No Question game (tracker)


teolemon:
## Still to be done
- [x] Images of the products should be aligned, possibly in a grid, possibly minimized ?
  - [ ] <img width="426" alt="image" src="https://user-images.githubusercontent.com/1689815/177146401-789f7b57-2eda-4872-8d58-866086d3f0a4.png">
- [x] Buttons should be centered and separated by a few horizontal pixels in mobile mode
  - [ ] 
- [ ] View should be centered in mobile mode
  - [ ] #101
  - [ ] No question remaining
    - [ ] #102
- [x] #386
- [x] Remaining >> Remaining questions
- [ ] Allow to type a value by adding a field (ideally with autocompletion)
- [x] Honor English shorcuts for English
- [ ] ![Image](https://user-images.githubusercontent.com/1689815/177634806-68bae3bc-d052-4509-bac9-c7149b5a7a53.png)
- [ ] Add a view with only products which don't have a category at all (states=categories-to-be-completed) for the category game

---

## Done - Category/Brand game
- [x] The Hunger Games logo is not aligned with the menus in the top bar
  - [ ] <img width="706" alt="image" src="https://user-images.githubusercontent.com/1689815/177146513-265ea486-6240-4f3a-9e50-86e51b2d7ba5.png">
- [x] Add padding after "See examples of this brand"
  - [ ] <img width="347" alt="image" src="https://user-images.githubusercontent.com/1689815/177146573-788cd6f0-607a-4694-8b5c-22da329c8380.png">
- [x] View and Edit button should be augmented with icons, and possibly be turned into buttons
  - [ ] <img width="347" alt="image" src="https://user-images.githubusercontent.com/1689815/177146705-29b2bb57-8f79-4c45-970f-ec03176fa9e7.png">
- [x] brand / category / label dropdown should be more explicit "Only show brands" "Only show categories"
    - [x] Offer to remove value when no question left for this value <img width="693" alt="image" src="https://user-images.githubusercontent.com/1689815/177630690-cb21aafd-6767-405d-b758-fff33e03d6fd.png">

### Part of
- #37

teolemon:
yep, can you comment on this issue, so that I can assign it to you ?
https://github.com/openfoodfacts/hunger-games/issues/386

Apurv428:
@teolemon can you give more details about the issue: Images of the products should be aligned, possibly in a grid, possibly minimized?
I'm interested in working on it.

[...]

PR #459, opened 2023-01-05
fix: 

Apurv428:
### What
- Issue #458: Updated the styles of the buttons 

### Fixes bug(s)
- 

### Part of 
-

### DIFF
--- a/src/pages/questions/QuestionDisplay.jsx
+++ b/src/pages/questions/QuestionDisplay.jsx
@@ -10,6 +10,7 @@ import Typography from "@mui/material/Typography";
 import LinkIcon from "@mui/icons-material/Link";
 import DeleteIcon from "@mui/icons-material/Delete";
 import DoneIcon from "@mui/icons-material/Done";
+import QuestionMarkIcon from '@mui/icons-material/QuestionMark';
 import MuiLink from "@mui/material/Link";
 import { useTheme } from "@mui/material/styles";
 import useMediaQuery from "@mui/material/useMediaQuery";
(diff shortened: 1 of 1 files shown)


alexfauquette:
I'm not sure about what was expected from this issue, maybe better to clarify what has to be done before writing the PR.


A screenshot is not an issue, since we can not know if it's the goal or if it highlights a bug. Maybe @teolemon will have more feedback

braaar:
The linked issue is a bit poorly formulated, indeed, but I think these buttons look better than the current buttons, at least. The current buttons are way too big. I don't see the point of that circular highligh on the skip button, though.

alexfauquette:
> current buttons are way too big

Even on mobile? I tend to prefer big button on mobile to be sure to do not miss click

image

braaar:
> > current buttons are way too big
> 
> Even on mobile? I tend to prefer big button on mobile to be sure to do not miss click
> 


That's fair. I was more thinking for desktop. The buttons are simply ginourmous. I want to make more space for the image by shrinking them, as I outline in my discussion post #463

Apurv428:
@alexfauquette

Apurv428:
@alexfauquette initially the buttons were aligned like this:
Screenshot_20230107_112209

But as mentioned in issue #251, it was asked to align the buttons in a more compact manner. So working on it, the following is the way of how the buttons are now placed in:

## Wide Screen View (Computer):
Screenshot_20230107_112645

## Mobile View:
Screenshot_20230107_112705

alexfauquette:
In the issue, it's a screenshot taken by @teolemon (not a figma design) So I assume it was the initial design when I put 3 buttons to do the job without thinking about how they should be placed.

**on desktop**, I agree it's an improvement to save some vertical place to allow the image to be more visible 👍

**on mobile** I tend to disagree it's an improvement for two reasons:

1. The previous one is more compact and on mobile, it's even more critical to save space for the image. The initial idea behind hunger-games is to allow contributors to annotate products instead of scrolling Twitter in the train (mobile first).
2. The previous one is more semantic: Your thumb is up ready to answer: left or right depending on yes or no, bottom if you don't know. Yes or No are similar (you answer the question) and distinct from skip. When I play I gave me 2 seconds to press left or right. If I have a doubt I skip

braaar:
I agree, stacking them all vertically on mobile is a mistake.

PR #464, opened 2023-01-07
more logos

teolemon:
### What
- more logos in more thematic categories


VaiTon:
Shouldn't the images be `static.openfoodfacts.org/...`?

alexfauquette (approved):

PR #476, opened 2023-01-14
Fix: Autocomplete behavior

alexfauquette:
If the value is not taken from the taxonomy it add a warning:

image

Otherwise it just show the associated tag

image

The `blur` event listener was buggy. Now it should not be necessary anymore to press enter. leaving the field is enought to update the value


teolemon:
We could add a link on how to edit the taxonomy

teolemon (approved):

Issue #483, opened 2023-01-17
No intuitive way to close a modal on mobile


teolemon:
### What

- No intuitive way to close a modal on mobile (you have to click outside)
- Can we have a modal control (eg a X) ?

![image](https://user-images.githubusercontent.com/1689815/212879660-cce0e799-563b-491b-85ba-6d7f3a85c94c.png)

Sudhanva-Nadiger:
Can I work on this ?

alexfauquette:
Yes, feel free.

I would add that on mobile the modal could be full screen. It does not make a lot of sens to lose pixels around

Sudhanva-Nadiger:
@alexfauquette have sent [pr](https://github.com/openfoodfacts/hunger-games/pull/495) for this issue. Thanks for the suggestion, done as u said!

PR #495, opened 2023-01-27
fix: add close button to close modal

Sudhanva-Nadiger:
in small devices, it was nonintuitive to close the modal, but now it's solved using the close icon, onTappig it will close.

### What
- The closing modal was hard in small devices, it's now solved by adding a close modal button, which on tap closes the modal
- I kept the close button left so that if the image overflows on the right side also the close button would be there

### Fixes bug(s)
-  #483


alexfauquette:
Sounds good, I just added some extra stuff:

- Put the cross on the right (it's not impacted by the scroll)
- I removed the scroll. You can already zoom in on the image with your mouse
- When you put an Icon, it's better to wrap it into an `<IconButton/>` that is responsible of the `onClick` listener (it add some better interaction, such as focussing, pointer modification on hover ...)

image


Are those modifications ok for you?

Sudhanva-Nadiger:
Yesssss ! I will do that !

Sudhanva-Nadiger:
Hi @alexfauquette in a small screen we can make height also to fill in the screen, else the image will only occupy half the spaces, 

https://user-images.githubusercontent.com/93595710/215250238-deef77b4-7a78-493c-a78a-4ab8134c6e8a.mp4

alexfauquette (approved):

alexfauquette:
> we can make height also to fill in the screen

Not a big fan since it adds distortion to the image

PR #498, opened 2023-01-30
feat: add loading animation before question display

### LINKED ISSUE #496: Add loading animation when user goes to question route
### Problem
When a user goes to the question tab currently it shows only "loading" text.

### Proposed solution
instead we can add a loading animation 

### Mockups
image

Sudhanva-Nadiger:
Fix #496

- before it was showing only loading text, now it shows loading circular animation

### Fixes bug(s)
- #496

### DIFF
--- a/src/i18n/common.json
+++ b/src/i18n/common.json
@@ -114,7 +114,8 @@
     "log_in": "Login",
     "sign_up": "Sign up",
     "login_title": "Make your annotation count",
-    "login_description": "Currently you are not logged in Open Food Facts. To get your answers directly applied to products, and associated to your account, consider logging in to your Open Food Facts account, or creating one"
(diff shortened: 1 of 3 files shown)


alexfauquette (approved):
Just did few modifications:

```diff
width: 100vw;
width: 100%;
```

Overwise your div 100vw plus the container padding leads to a horizontal scroll bar

---

Moved the import. I try to group by internal/library import to keep clear from where stuff are coming from

---

I know it's not in issue template, but a nice think is to add a line `Fix #<the number of the issue fixed>`

It links your PR and the issue, such that merging the PR closes automatically issue.

Otherwise, we tend to keep open issues that are already fixed

more details here: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

Sudhanva-Nadiger:
Oh okay , sorry for that ! I will fix the template!!

Issue #386, opened 2022-12-02
Capitalize countries


teolemon:
### What
- Capitalize countries

### Part of
- #37

vedant-z:
@teolemon I just need some help.

Do I need to update the countries.json or other file?

teolemon:
- It's currently here: https://github.com/openfoodfacts/hunger-games/blob/master/src/components/QuestionFilter/const.js#L12
- We also have this: https://github.com/openfoodfacts/hunger-games/blob/master/src/assets/countries.json

alexfauquette:
For lazy people, capitalizing can also be done by using CSS 😇
https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform

vedant-z:
Even the most workaholic man on this planet will not be able to capitalise all the countries name manually @alexfauquette 

Thanks for helping 😊

vedant-z:
I had one doubt though. Should I capitalise the id as well in json file, because the label is already capitalised.

alexfauquette:
You should not touch to ids in any case, since they are the keys used to fetch the API. Modifying ids implies breaking API call

If we are talking about the country input in the question game filter, the options are defined as follow in `QuestionFilter.jsx`
```jsx
options={countryNames}
getOptionLabel={(countryTag) => countryTag.slice(3)}
```

For example when `countryTag="en:france"` it shows the option `"france"`.

Two solutions:
- Using https://github.com/openfoodfacts/hunger-games/blob/master/src/assets/countries.json to get a stardised name from the id
- Improve `(countryTag) => countryTag.slice(3)` such that it capitalize the name

StarTrooper08:
Hey @teolemon, Is this issue still open?

Or is it solved?

I'm just getting started with contributing to Open Food Facts projects.
And I think this issue will be great for getting starting.

Please let me know if this issue is it still open to work on!?

Sudhanva-Nadiger:
I solved the issue, if u need any help let me know @vedant-z

alexfauquette:
@Sudhanva-Nadiger If you already fixed it feal free to open a PR

Sudhanva-Nadiger:
Okay !! Will do that !!

StarTrooper08:
I think this issue has PR merged and it can be closed now!!

PR #504, opened 2023-01-31
fix: capitalize country names.

Sudhanva-Nadiger:
### What
- Country names were in small letters
- Now the first letter has been capitalized 

### Fixes bug(s)
- #386 

### Part of 
- #37


alexfauquette:
`id`s can not be overridden. It's the key used to communicate with servers. Same for `countryNames`. We need the `'en:....'` to know what is the standard name of the country

The trick is to distinguish e value and the displayed text: https://github.com/openfoodfacts/hunger-games/issues/386#issuecomment-1335476404

Sudhanva-Nadiger:
ohh yeah yeah got it!!! Like this ?
image
but if we type anything not in the list it shows no options noe

alexfauquette:
> but if we type anything not in the list it shows no options noe

I don't get your point. If you update the PR, I could have a look to the behavior

Sudhanva-Nadiger:
image

#
This is what i was asking ! in the previous comment. If we type anything which is not in country names it shows no options !

alexfauquette:
Make sense that it does not show options if it's not part of the options. We put countries of the first countries concerned by eco-score. But you can use [`countries.json`](https://github.com/openfoodfacts/hunger-games/blob/c530360ad0b0e9475b49fcbfa30bec1b85ad39cb/src/assets/countries.json?rgh-link-date=2022-12-02T16%3A07%3A03Z#L643-L648) to get all the countries with product in OFF

Sudhanva-Nadiger:
> Make sense that it does not show options if it's not part of the options. We put countries of the first countries concerned by eco-score. But you can use [`countries.json`](https://github.com/openfoodfacts/hunger-games/blob/c530360ad0b0e9475b49fcbfa30bec1b85ad39cb/src/assets/countries.json?rgh-link-date=2022-12-02T16%3A07%3A03Z#L643-L648) to get all the countries with product in OFF

Yeah, but we are currently using some of the country names, should we provide all countries as options in the card?

Sudhanva-Nadiger:
Hey I have opened a new [pr](https://github.com/openfoodfacts/hunger-games/pull/522), this window whitelisted file names giving me lots of trouble, so had to do lots of branch lol and send pr.

PR #522, opened 2023-02-01
fix: capitalise country name.

Sudhanva-Nadiger:
### What
- Capitalise country name while showing

### Screenshot
image
#
image

### Fixes bug(s)
- 

### Part of 
- #37

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -44,7 +44,9 @@ export function sleep(ms) {
 
 //to provide capitalised country name; en:france => France
 export const capitaliseName = (string) => {
-  if (!Boolean(string)) return;
+  if (!string) {
+    return string;
+  }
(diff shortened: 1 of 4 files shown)


Sudhanva-Nadiger:
So this is all capitalisation we wanted ?? 😁

src/utils.js
```diff
@@ -43,7 +43,8 @@ export function sleep(ms) {
 }
 
 //to provide capitalised country name; en:france => France
-export const capitaliseName = (country) => {
-  let name = country.slice(3);
+export const capitaliseName = (string) => {
+  if (!Boolean(string)) return;
```
alexfauquette, line 47:
```suggestion
  if (!string) {
    return string;
  }
```

PR #614, opened 2023-05-18
Allow the crowdin workflow to be run on-demand

VaiTon:


teolemon (approved):
Looks good to me but we can only do a synchronisation at one time for the open food facts organization for all repositories, and that many hours are already taken

VaiTon:
@teolemon can we sync just once a day? With this addition the periodic sync is still in place, but we can also trigger it manually

teolemon (approved):

PR #641, opened 2023-06-27
Fix: fixed warn on empty values

### LINKED ISSUE #480: Don't warn on empty value
image

Mount-Blanc:
### What
In the ternary operator, i added two && checks that will be true only for non empty string and not null.  This fixed the issue but caused the label filter to lose focus once the user types in a letter. So,additionally, I added a useEffect and useRef hook.  The useEffect hook re-runs everytime the innerValue changes. The useRef hook holds on to the value that the user types and re-focuses on the label filter, this runs everytime the useEffect hook runs.

### Screenshot
Screen Shot 2023-06-27 at 2 01 47 PM

### Fixes bug(s)
Fix #480

### DIFF
--- a/src/components/LogoSearchForm.jsx
+++ b/src/components/LogoSearchForm.jsx
@@ -1,4 +1,5 @@
 import * as React from "react";
+import { useEffect, useRef } from "react";
 
 import Stack from "@mui/material/Stack";
 import TextField from "@mui/material/TextField";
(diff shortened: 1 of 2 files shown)


alexfauquette:
Thanks for the contribution. You're losing focus because you are switching between those `TextFIeld` and `LabelFilter`.

What matter in this issue is the warning, not the field itself. It would be easier to go into the definition if the `LabelFilter` and here modify the `helperText` prop.

By the way, I noticed the `<LabelFilter />` has a prop `showKey` that you can turn to false to hide this helper text

Mount-Blanc:
Wow you're right! Turning the showKey false completely resolved the problem. It seems I over-engineered the solution, thanks for letting me know !

alexfauquette:
If you put `show={false}` it will never show the warning

alexfauquette (approved):
Thanks for the contribution

PR #657, opened 2023-07-07
Remove webpack bundling

VaiTon:
### What
- AFAIK, we don't need to bundle everything in a single JS file, as the app is deployed separately from the main website.


VaiTon:
@stephanegigandet @alexgarel I'm tagging you since the bundle was made to "facilitate integration in OFF main site".

alexfauquette (approved):

alexgarel:
Thanks @VaiTon 

I think you can remove this indeed unless @teolemon has some information I'm not aware off.

PR #664, opened 2023-07-09
Update to yarn v2, without pnp

VaiTon:
https://yarnpkg.com/getting-started/qa#why-should-you-upgrade-to-yarn-modern


alexfauquette (approved):
Good for me

PR #663, opened 2023-07-08
Use Vite instead of CRA

VaiTon:
### What
- Use Vite instead of Create React App for faster dev iteration.

### Depends on
- https://github.com/openfoodfacts/hunger-games/pull/656

PR #666, opened 2023-07-09
Fix GitHub pages deploy

VaiTon:
- In #663 I forgot to modify the `build.sh` script. This fixes it.
- Now we're directly pushing to GitHub Pages with an action, instead of pushing to the branch.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -36,9 +36,12 @@ jobs:
       - name: Deploy
         run: yarn deploy
 
+      - name: Upload artifact
+        uses: actions/upload-pages-artifact@v1
+        with:
+          path: 'dist'
+
       - name: Deploy to github pages 🚀
         if: github.ref == 'refs/heads/master'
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -36,9 +36,12 @@ jobs:
...
+          path: 'dist'
+
       - name: Deploy to github pages 🚀
         if: github.ref == 'refs/heads/master'
-        uses: JamesIves/github-pages-deploy-action@v4.4.2
-        with:
-          branch: gh-pages # The branch the action should deploy to.
-          folder: build # The folder the action should deploy.
+        id: deployment
+        uses: actions/deploy-pages@v2
```
alexfauquette, line 47: I'm not sure to understand, the difference between the two

"Upload artifact" is uploading the `dist` folder. But then what is doing the `actions/deploy-pages@v2`?

alexfauquette: Thanks for the explaination

alexfauquette (approved):

PR #669, opened 2023-07-10
aligned loader center while loading pages

saikiranpatil:
### What
Center align the loader

### Screenshot
Capture


alexfauquette (approved):

VaiTon:
Also, It would be better to create a component and use it everywhere instead of implementing the same one in multiple places.

VaiTon:
@saikiranpatil did you mean to close this?

PR #675, opened 2023-07-11
Since all routes are imported directly in app.jsx, This file is of no…

PrateekBhat2:
### What
- Since all routes are imported directly in app.jsx, this file is of no use


VaiTon (approved):

VaiTon:
Thank you!

PR #676, opened 2023-07-11
chore: utils.js to TypeScript

VaiTon:


alexfauquette (approved):
If typescript does not complain, feel free to merge such PR without review :)

PR #681, opened 2023-07-14
chore: remove react-scripts dependency

VaiTon:
As we're now using Vite, we don't need that dependency.


alexfauquette (approved):
Nice, it will remove some false positives about security

PR #685, opened 2023-07-24
chore: convert pages to TypeScript

VaiTon:
### What
- 2nd PR to convert the project to TypeScript

### How to review
You should view the PR commit by commit. In that way it should be easier reviewing.


src/reportWebVitals.ts
```diff
@@ -0,0 +1,15 @@
+const reportWebVitals = async (
```
alexfauquette, line 1: Personally I never used this file 🙈

alexfauquette (approved):

PR #708, opened 2023-08-02
few bug fixes

alexfauquette:
Fixes per commit:

- Impossible to switch pages in the tour (app crashes). The reason being the asset import not supported by viteJs
- pretty-quick does not support prettier v3. So I downgrade prettier to v2 otherwise husky prevent me from commiting(the [GitHub issue](https://github.com/azz/pretty-quick/issues/164#issuecomment-1625533955))
- The loader background was hardcoded with a dark color, leading to a black spinner on a black background in light mode. Here is the before after 🙈

image
image

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -10,7 +10,7 @@ on:
 
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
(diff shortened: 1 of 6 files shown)


teolemon (approved):
Great 🎉

PR #713, opened 2023-08-10
feat: Allow to remove some logos from deep search

alexfauquette:
Pressing remove will remove selected logos from the list. No request is sent to the server

image

### DIFF
--- a/src/pages/logos/LogoDeepSearch.jsx
+++ b/src/pages/logos/LogoDeepSearch.jsx
@@ -284,6 +284,10 @@ export default function LogoSearch() {
     );
   };
 
+  const removeSelectedLogos = () => {
+    setLogosToAnnotate((prev) => prev.filter((logo, index) => !logo.selected));
+  };
+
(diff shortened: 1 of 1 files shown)


VaiTon:
Since they are not really removed could we rename the button to "Hide"?

VaiTon (approved):
LGTM apart from the comment above!

alexfauquette:
> Since they are not really removed could we rename the button to "Hide"?

From the point of view of the web page, they are removed.

The app has two arrays:
- `annotatedLogos`: The logos that are annotated and that we can use to fetch more logos
- `logosToAnnotate`: The logo fetched by the API that needs to be annotated

So this function removes those logos from the app memory
```jsx
const removeSelectedLogos = () => {
    setLogosToAnnotate((prev) => prev.filter((logo, index) => !logo.selected));
  };
```

Is their something in particular that let you say they are just hidden?

VaiTon:
"Remove" made me think about "remove logo from the server", but I'm ok with either

PR #742, opened 2023-09-22
fix: Avoid loading logos without a filtering defined

### LINKED ISSUE #739: Don't perform logo search by default on logo search page
On https://hunger.openfoodfacts.org/logos/search, the following HTTP request is sent automatically: https://robotoff.openfoodfacts.org/api/v1/images/logos/search/?count=25&random=true

This is a very heavy request for the DB, as , as we perform a double join over 2 tables to get the number of items without any filtering. It would be better not to send any request when the page load, and wait for the user to provide input.

alexfauquette:
Fix #739

### DIFF
--- a/src/pages/logos/LogoSearch.jsx
+++ b/src/pages/logos/LogoSearch.jsx
@@ -41,7 +41,7 @@ const request = async ({ barcode, value, type, count }) => {
 export default function LogoSearch() {
   const { t } = useTranslation();
 
-  const [isLoading, setIsLoading] = React.useState(true);
+  const [isLoading, setIsLoading] = React.useState(false);
   const [searchState, setSearchState] = useUrlParams(
     {
       type: "",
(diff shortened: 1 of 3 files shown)


src/pages/logos/LogoSearch.jsx
```diff
@@ -61,7 +61,16 @@ export default function LogoSearch() {
...
   );
 
+  const filterStateHasValue =
+    (searchState.type &&
+      (TYPE_WITHOUT_VALUE.includes(searchState.type) || searchState.value)) ||
+    searchState.barcode;
+
   React.useEffect(() => {
+    if (!filterStateHasValue) {
+      // Avoid fetching data is no valiue to filter
```
teolemon, line 71:
```suggestion
      // Avoid fetching data if no value to filter
```

src/pages/logos/ProductLogoAnnotations.jsx
```diff
@@ -87,6 +93,12 @@ const useLogoFetching = (filter) => {
   }, [filter]);
 
   React.useEffect(() => {
+    const filterStateIsIncomplet = !filter.tagtype || !filter.tag;
+    if (filterStateIsIncomplet) {
+      // Avoid fetching data is no valiue to filter
```
teolemon, line 98:
```suggestion
      // Avoid fetching data if no value to filter
```

Issue #688, opened 2023-07-25
Bug: white screen in Logos search when clicking on value field first


DanieliusAsm:
### What
White screen when clicking on value field first. You have to click on it first while not having anything in Type selected.

This bug only happens on logos that have no annotation.

### Steps to reproduce the behavior:
1. Go to https://hunger.openfoodfacts.org/logos/search?barcode=4770179104378 (or any other link)
3. Click on value (dont select type leave it blank)
4. Type in something

Im on chrome browser if thats needed.

PrateekBhat2:
can i work on this ?

alexfauquette:
Yes, the solution might be to disable the field

PrateekBhat2:
@DanieliusAsm I am unable to reproduce this even after following the procedure mentioned. Can you elaborate more?

alexfauquette:
Here is the reproduction

after clicking value I'm pressing a letter, leading to the white screen

![image](https://github.com/openfoodfacts/hunger-games/assets/45398769/15b57ae5-3741-4271-bd8e-d7981667d064)

PrateekBhat2:
thanks! will work on it

PR #764, opened 2023-10-01
Fix : Blank screen problem

PrateekBhat2:
### What
- Prevents the reading properties on null error using optional chaining
- ToLowerCase function execution will be prevented when value of 'value' is equal to null.

### Fixes bug(s)
- fix: #688


alexfauquette (approved):
Very nice. We should go to typescript to prevent those kind of bugs :)

PR #765, opened 2023-10-04
Let every one access to the logo dashboard

### LINKED ISSUE #716: Make the logo dashboard link in Hunger Games menu visible to all users
This was hidden by default as this was a test, but I think it would be valuable to display it to all users.
We can also clean the settings, "show logo page" looks like a legacy setting to me (it's always displayed, isn't it)?

### Part of
- #252

alexfauquette:
Fix #716


teolemon:
Thanks @alexfauquette 
Merging

PR #850, opened 2024-01-02
core: use node SDK

alexfauquette:


teolemon:
That's a great idea ! @alexfauquette @VaiTon

teolemon:
still a great idea, but merge conflict :-/ @alexfauquette

teolemon (approved):
Nice. We should dogfood our SDKs 👍

alexfauquette:
@VaiTon Abandoned: I can close it if you want.

More precisely, I still think it's an important move, but I don't have bandwidth for now, so I will try back later

teolemon:
@alexfauquette I renamed your PR since we have renamed the SDK to reflect that it is in fact a JavaScript SDK.

PR #892, opened 2024-02-19
feat: add country list in settings page, using it as default country in eco-score

### LINKED ISSUE #887: Save user country country
Followup of #879 (@teolemon, @benbenben2  if you have an opinion on the behavior)

In `/settings` we allow to set the language but not the country.

image

The idea would be to to save in the localeStorage the user country.

By default country should not be defined (equivalent of being set to world)

If a default country is set, it should be used in most games. Especially in the eco-score dashboard

@Gaurav-Verma07 you might be interested by this follow up

Gaurav-Verma07:
…in eco-score

### What
Add country drop-down in settings page using `countries.json`.
Settings it as default in `eco-score` page.

### Screenshot
image

### Fixes bug(s)

Fix #887

### DIFF
--- a/src/pages/eco-score/index.tsx
+++ b/src/pages/eco-score/index.tsx
@@ -199,7 +199,7 @@ export default function EcoScore() {
         >
           {countryNames.map((country) => (
             <MenuItem value={country.id} key={country.id}>
-              {capitaliseName(country.id)}
+              {country.label}
             </MenuItem>
           ))}
         </TextField>
--- a/src/pages/eco-score/index.tsx
+++ b/src/pages/eco-score/index.tsx
@@ -11,7 +11,6 @@ import SmallQuestionCard from "../../components/SmallQuestionCard";
 import Opportunities from "../../components/Opportunities";
 import { DEFAULT_FILTER_STATE } from "../../components/QuestionFilter/const";
 import { useTranslation } from "react-i18next";
-import { capitaliseName } from "../../utils";
 import Loader from "../loader";
 import { useSearchParams } from "react-router-dom";
 import { localSettings } from "../../localeStorageManager";
--- a/src/pages/eco-score/index.tsx
+++ b/src/pages/eco-score/index.tsx
@@ -157,7 +157,7 @@ export default function EcoScore() {
   const localData= localSettings.fetch();
   const [searchParams, setSearchParams] = useSearchParams();
   const [selectedCountry, setSelectedCountry] = React.useState(
-    searchParams.get("cc") || localData['country'] || countryNames[0],
+    searchParams.get("cc") || localData['country'] || countryNames[81].id,
   );
 
   React.useEffect(() => {
--- a/src/pages/eco-score/index.tsx
+++ b/src/pages/eco-score/index.tsx
@@ -157,7 +157,7 @@ export default function EcoScore() {
   const localData= localSettings.fetch();
   const [searchParams, setSearchParams] = useSearchParams();
   const [selectedCountry, setSelectedCountry] = React.useState(
-    searchParams.get("cc") || localData['country'] || countryNames[81].id,
+    searchParams.get("cc") || localData['country'] || "en:france",
   );
 
   React.useEffect(() => {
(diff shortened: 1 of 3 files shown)


src/pages/eco-score/index.tsx
```diff
@@ -148,33 +150,19 @@ const ecoScoreCards = [
...
-];
+
 
 export default function EcoScore() {
   const { t } = useTranslation();
+  const localData= localSettings.fetch();
   const [searchParams, setSearchParams] = useSearchParams();
   const [selectedCountry, setSelectedCountry] = React.useState(
-    searchParams.get("cc") || countryNames[0],
+    searchParams.get("cc") || localData['country'] || countryNames[81].id,
```
alexfauquette, line 160:
```suggestion
    searchParams.get("cc") || localData['country'] || "en:france",
```

Effectively we don't have a lot of users in Afghanistan (for now 😁)
Using directly the id might be more robust. Those ids should not change

src/pages/eco-score/index.tsx
```diff
@@ -148,33 +150,19 @@ const ecoScoreCards = [
...
   );
 
   React.useEffect(() => {
     setSearchParams({ cc: selectedCountry });
-  }, [selectedCountry]);
+  }, [selectedCountry, searchParams]);
```
alexfauquette, line 165: I don't see why adding this dependency

```suggestion
  }, [selectedCountry]);
```

src/pages/eco-score/index.tsx
```diff
@@ -210,8 +198,8 @@ export default function EcoScore() {
           sx={{ width: 200 }}
         >
           {countryNames.map((country) => (
-            <MenuItem value={country} key={country}>
-              {capitaliseName(country)}
+            <MenuItem value={country.id} key={country.id}>
+              {capitaliseName(country.id)}
```
alexfauquette, line 202: What about using the label directly?
```suggestion
              {country.label}
```

src/pages/settings/index.jsx
```diff
@@ -22,15 +22,18 @@ import DevModeContext from "../../contexts/devMode";
...
 export default function Settings() {
   const { t, i18n } = useTranslation();
   const theme = useTheme();
   const colorMode = React.useContext(ColorModeContext);
-
+  const localData= localSettings.fetch();
   const [language, setLanguage] = React.useState(i18n.language);
   const { devMode, setDevMode, visiblePages, setVisiblePages } =
     React.useContext(DevModeContext);
+  const [selectedCountry, setSelectedCountry]= React.useState(localData['country'])
```
alexfauquette, line 37:
```suggestion
  const [language, setLanguage] = React.useState(i18n.language);
  const { devMode, setDevMode, visiblePages, setVisiblePages } =
    React.useContext(DevModeContext);
  const [selectedCountry, setSelectedCountry]= React.useState(() => localSettings.fetch()['country'])

```

To avoid fetching localeStorage every render whereas we only need it at initialization

src/pages/settings/index.jsx
```diff
@@ -71,6 +79,19 @@ export default function Settings() {
...
+        <TextField
+          select
+          label={t("eco-score.countryLabel")}
+          value={selectedCountry}
+          onChange={handleCountryChange}
+          sx={{ width: 200 }}
+        >
+          {countryNames.map((country) => (
+            <MenuItem value={country.id} key={country.id}>
+              {capitaliseName(country.id)}
```
alexfauquette, line 91:
```suggestion
              {country.label}
```

alexfauquette (commented):
Sounds good. I added some minor comments

src/pages/eco-score/index.tsx
Gaurav-Verma07: > I don't see why adding this dependency

Actually it's kind of important, if you go to https://hunger.openfoodfacts.org/eco-score?cc=en%3Afrance
and click `eco-score` in navbar, the query parameter disappears which is not an expected behavior. 
What you think?

alexfauquette: ok, why not :+1:

alexfauquette (approved):

PR #1002, opened 2024-07-22
Fix robotoff API call for countries

### LINKED ISSUE #1001: Use parameters countries= when fetching insights
Country filtering does not work any longer (ex: https://hunger.openfoodfacts.org/questions?type=label&country=en%3Afrance), because `country=` parameter is not working any longer for the following endpoints:
- `/questions/random`
- `/questions/popular`
- `/questions/unanswered`
-  `/questions`
- `/insights/random`
- `/insights`
The `countries` parameter should be used instead with a two-letter code (it accepts several countries).
I thought I had created an issue on Hunger Games, but I obviously didn't, sorry for that! 

And by the way, `/questions/random` and `/questions/popular` are deprecated in favor of `/questions?order_by={random,popularity}`:
https://openfoodfacts.github.io/robotoff/references/api/#tag/Questions/paths/~1questions/get

alexfauquette:
Fix #1001

Use an intermediate function to convert id (en:France) to code (FR)

~@raphael0202 is it case-sensitive?~

After trying, it is 🙈 so I lowercased everything

For example, 

https://deploy-preview-1002--warm-cannoli-79bbb2.netlify.app/questions?country=en%3Aaustria

is calling

https://robotoff.openfoodfacts.org/api/v1/questions/?countries=at&order_by=popularity&lang=en&count=25&page=1

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,32 +1,32 @@
 # Add labels to any any pull request with changes to the specified paths
 Github Actions:
-- changed-files:
-  - any-glob-to-any-file: '.github/**/*'
+  - changed-files:
+      - any-glob-to-any-file: ".github/**/*"
 
 Questions game:
-- changed-files:
-  - any-glob-to-any-file: 'src/pages/questions/**/*'
+  - changed-files:
+      - any-glob-to-any-file: "src/pages/questions/**/*"
 
 🖼️ Logo game:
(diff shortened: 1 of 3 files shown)


teolemon (approved):
Thanks @alexfauquette 
Given the impact, I'm speedy merging 👍

PR #1056, opened 2024-10-12
Update z.json

krz:
### What
Add some missing German translations and fix some typos

### DIFF
--- a/public/data/de/category/z.json
+++ b/public/data/de/category/z.json
@@ -5,11 +5,11 @@
   },
   {
     "key": "en:brown-sugars",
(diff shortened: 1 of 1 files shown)


alexfauquette:
This fils is generated by running the `update-taxonomy-suggestions.js`. If we merge this PR it will be overridden the next time the script get run.

If you know how to do, please run the script for german locale.

If all typo are not fixed, you can fix them in the taxonomy (should be in [this repo](https://github.com/openfoodfacts/openfoodfacts-server)) which is our source of truth. Such that all the projects get the fix

PR #1111, opened 2024-12-17
Make nutrition table visible in the menu

alexfauquette:
This PR replaces hardcoded text by localized text. For now only EN is available.

And it adds a link in the menu.

Like logos games, if user is not logged in they will see a page requesting to logg-in


teolemon (approved):
Fair enough. I've spotted that the global country switcher has probably broken the Eco-Score game

Issue #740, opened 2023-09-19
3 butter chickens in country UK, 0 without country filter


teolemon:
### What
- 3 butter chickens in country UK, 0 without country filter
- https://hunger.openfoodfacts.org/questions?country=en%3Auk&type=category&value_tag=en%3Abutter-chicken
- Unsure if frontend (Hunger Games) or backend issue (Robotoff). EDIT: definitely frontend, loading immediately the URL without filter works fine

alexfauquette:
Because the front-end assume uk code is `en:united-kingdom` and the backend `en:uk`

alexfauquette:
@teolemon  I was going to replace `en:united-kingdom` by `en:uk` when I realized both exist:

For example in agribalise campaign about the chicken category, we have: 
- 238 questions with united-kingdom: (see [here](https://hunger.openfoodfacts.org/questions?value_tag=en%3Achickens&type=category&campaign=agribalyse-category&country=en%3Aunited-kingdom&sorted=true))
- 2232 questions with united-kingdom: (see [here](https://hunger.openfoodfacts.org/questions?value_tag=en%3Achickens&type=category&campaign=agribalyse-category&country=en%3Auk&sorted=true))

Maybe there is something to do on OFF or Robotoff to have a unique id, or I'm missing a subtility between those two @raphael0202

alexfauquette:
Looks like it's a fixed issue

https://hunger.openfoodfacts.org/questions?type=category&value_tag=en%3Abutter-chicken
https://hunger.openfoodfacts.org/questions?country=uk&type=category&value_tag=en%3Abutter-chicken&sorted=true

PR #743, opened 2023-09-22
fix: change the id from en:united-kingdom to en:uk

alexfauquette:
- Fix #740

PR #1136, opened 2025-02-04
fix(#740): change the id from en:united-kingdom to en:uk

nmetifiot:
It seems that this piece of code is missing in the final version

### DIFF
--- a/src/assets/countries.json
+++ b/src/assets/countries.json
@@ -1405,7 +1405,7 @@
     "countryCode": "ae"
   },
   {
-    "id": "en:united-kingdom",
+    "id": "en:uk",
     "label": "United Kingdom",
     "languageCode": "en",
     "countryCode": "uk"


alexfauquette:
> this piece of code is missing

Could you explain why?

nmetifiot:
In this issue: https://github.com/openfoodfacts/hunger-games/issues/740 you made this PR: https://github.com/openfoodfacts/hunger-games/pull/743 
But it was not merged. It seems that there is another fix for this issue but there is still this piece of code that as not been integrated.
I proposed this PR just if it is an oversight.

alexfauquette:
I don't fully understood that issue. I'm wondering if it's an error on our side or if their is some inconsistencies between OpenFoodFacts and robotoff. Maybe you better know the issue than me who only use the french version

What I'm sure is that we can not merge this PR because this file is generated by `update-countries.js` ;)

VaiTon:
@alexfauquette @nmetifiot I would close this PR for the time being if you're ok with that

Issue #1176, opened 2025-05-21
Menu doesn't close after clicking on a link


raphael0202:
Bug report from a Hunger Game user:

> I've just started playing around with the hunger games a bit and noticed that the burger menu doesn't close again after I opened it to select my game, even tho it directs me to the correct site - it just stays on the screen, after reloading the page it's gone again. Tapping the burger sign again also doesn't do anything, reloading is the only thing that helps. I've attached a screen record of the issue to show you what I mean.

https://github.com/user-attachments/assets/686e6665-9cb8-4594-98bb-c36fc9dd1d86

alexfauquette:
On mobile, the navigation menu does not close when navigating

https://github.com/user-attachments/assets/5f4a5f06-fa3c-43f3-8f10-501da2898099

PR #1214, opened 2025-07-09
Fixed: Menu not closing

MishraTanishq619:
### What

Core of this issue is that Menu.open depends on anchorEl.
I found that, even if "Menu.open" = false ; the Menu doesn't closes.
So, I am proposing to do manual conditional rendering of "Menu" component with the Menu.open attribute provided from MUI.
This simple fix is solving the issue.

### Screenshot

https://github.com/user-attachments/assets/e1d33eb6-7587-4d10-b11b-52c2d744a5a0

### Fixes bug(s)
- #1176

### DIFF
--- a/src/components/ResponsiveAppBar.jsx
+++ b/src/components/ResponsiveAppBar.jsx
@@ -171,10 +171,6 @@ const ResponsiveAppBar = () => {
       return isPageVisible(page);
     });
 
-  React.useEffect(() => {
-    console.log("Boolean(anchorElNav) : ", Boolean(anchorElNav));
-  }, [anchorElNav]);
-
   return (
     <AppBar
       position="static"
--- a/src/components/ResponsiveAppBar.jsx
+++ b/src/components/ResponsiveAppBar.jsx
@@ -171,6 +171,10 @@ const ResponsiveAppBar = () => {
       return isPageVisible(page);
     });
 
+  React.useEffect(() => {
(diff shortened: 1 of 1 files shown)


src/components/ResponsiveAppBar.jsx
```diff
@@ -201,106 +205,109 @@ const ResponsiveAppBar = () => {
...
-                  );
-                }
-                if (page.children) {
-                  return (
-                    <List
-                      component="div"
-                      disablePadding
-                      key={page.translationKey}
-                    >
+            {anchorElNav && (
```
MishraTanishq619, line 208: I added only this line

MishraTanishq619: I tried that too. But it broke the menu.

https://github.com/user-attachments/assets/fd24568b-c0ed-4ce6-b123-8493c9e1ffbe

src/components/ResponsiveAppBar.jsx
```diff
@@ -171,6 +171,10 @@ const ResponsiveAppBar = () => {
       return isPageVisible(page);
     });
 
+  React.useEffect(() => {
+    console.log("Boolean(anchorElNav) : ", Boolean(anchorElNav));
+  }, [anchorElNav]);
+
```
alexfauquette, line 177:
```suggestion
```

PR #1229, opened 2025-07-29
chore: remove any from reportWebVitals, migrate ResponsiveAppBar to TS

VaiTon:
Also makes the navbar assume the "mobile" setting when below the "lg" breakpoint, because links are overlapping otherwise.


alexfauquette (approved):
Thanks :)

I'm wondering if the `reportWebVitals.ts` does something. It might be here just because of the `create-react-app` and never get plugged to any server

PR #1249, opened 2025-08-22
chore: more conversions to typescript

VaiTon:
### What
- 

### Fixes bug(s)
- 

### Part of 
-


src/robotoff.ts
```diff
@@ -61,7 +61,19 @@ const robotoff = {
...
+  questions(
+    filterState: {
+      insightType: string;
+      brandFilter: string;
+      valueTag: string;
+      countryFilter: string;
+      sortByPopularity: boolean;
+      campaign: string;
+      predictor: string;
+    },
```
alexfauquette, line 73: I don't remember if we always provide the full filter params object

```suggestion
    filterState: Partial<{
      insightType: string;
      brandFilter: string;
      valueTag: string;
      countryFilter: string;
      sortByPopularity: boolean;
      campaign: string;
      predictor: string;
    }>,
```

src/externalApi.ts
```diff
@@ -2,11 +2,11 @@ import axios from "axios";
 
 export const addImageFlag = ({ barcode, imgid, url }) => {
   axios
-    .put(`https://amathjourney.com/api/off-annotation/flag-image/${barcode}`, {
-      mode: "no-cors",
-      imgid,
-      url,
-    })
+    .put(
+      `https://amathjourney.com/api/off-annotation/flag-image/${barcode}`,
```
alexfauquette, line 6: I forgot it. This was a nintial experiment. But my personall server is down since a year or two 🙃

It should be replaced by a call to nutripatrol

https://nutripatrol.openfoodfacts.org/api/docs#/default/create_flag_api_v1_flags_post

src/pages/ingredient-detection/index.tsx
```diff
@@ -1,14 +1,22 @@
...
   return (
-    <React.Suspense fallback={<Loader />}>
-      <Box sx={{ p: 2 }}>
+    <Suspense fallback={<Loader />}>
+      <Box
+        sx={{
+          p: 4,
+          display: "flex",
+          justifyContent: "center",
+          alignItems: "center",
```
alexfauquette, line 15: Not TS related. Is there a particular reason to move them?

alexfauquette (approved):

alexfauquette (approved):

PR #1346, opened 2025-12-04
chore: update react and dependencies

stephanegigandet:
warning: not tested


alexfauquette (commented):
That's a bit too many major upgrade for a single PR. There are no risk about server components, so we can close it and do the updates one y one latter

Issue #1040, opened 2024-09-30
Allow Opening the Full Quality Image for Questions


Naruyoko:
### Problem
You can click on the "zoom" icon at the corner of an image in _Questions_ to open it in a modal where you can zoom in to small information. However, it uses the same shrunken images already shown before, which compresses away details that are too small relative to the image size. This is particularly a problem for text, which are unreadable for large images even if the original image is clear.

### Proposed solution
Load the original image in the zoomed view. It could be behind a "load original" button so the full image is not fetched in cases that do not need it.

PR #1350, opened 2025-12-08
Fix: Allow Opening Full-Quality Image for Questions (#1040)

Ramneet04:
### What
- Fixed the full-resolution toggle button on ZoomableImage component.
- Button now overlaps the image correctly and shows a subtle background.

### Screenshot
image1_hunger_games
image2_hunger_games
image3_hunger_games

### Fixes bug(s)
- #1040

### Part of 
- #1040

### DIFF
--- a/src/components/ZoomableImage.tsx
+++ b/src/components/ZoomableImage.tsx
@@ -97,23 +97,17 @@ const ZoomableImage = ({
             position: "relative",
           }}
         >
-          <Box position="relative">
+          <Box>
             <TransformWrapper limitToBounds={false} ref={apiRef}>
               <TransformComponent>
                 <img
                   src={showFullResolution && srcFull ? srcFull : src}
                   alt=""
-                  loading="lazy"
                   style={{
-                    maxHeight: "80vh",
+                    maxHeight: "calc(100vh - 160px )",
                     maxWidth: "100%",
-                    width: "100%",
-                    height: "auto",
-                    objectFit: "contain",
                     transform: `rotate(${rotation * 90}deg)`,
                     transformOrigin: "center",
-                    display: "block",
-                    margin: "auto",
                   }}
                 />
               </TransformComponent>


src/components/ZoomableImage.tsx
```diff
@@ -72,41 +81,88 @@ const ZoomableImage = ({
...
+                  display: "block",
+                  margin: "auto",
                 }}
               />
             </TransformComponent>
           </TransformWrapper>
+          {srcFull && (
+            <Button
+              disabled={loadingFullResolution}
+              variant={useFullResolution ? "contained" : "outlined"}
```
alexfauquette: outlined for all looks better. Otherwise the button can disapear because of transparency

image

src/components/ZoomableImage.tsx
```diff
@@ -72,41 +81,88 @@ const ZoomableImage = ({
...
+              backgroundColor: "rgba(0, 0, 0, 0.6)",
+              zIndex: 10,
+            }}
+          >
+            <IconButton onClick={() => setIsOpen(false)}>
+              <CloseIcon />
+            </IconButton>
+            <Loader />
+          </Box>
+        )}
```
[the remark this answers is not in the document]
Ramneet04: Yeah,  the loader should be scoped to the dialog content rather than the header. I’ll move it so it overlays only the image container.

Ramneet04:
@alexfauquette please do check it once, and let me know if any further changes required.

src/components/ZoomableImage.tsx
```diff
@@ -72,41 +84,91 @@ const ZoomableImage = ({
...
+                  style={{
+                    maxHeight: "80vh",
+                    maxWidth: "100%",
+                    width: "100%",
+                    height: "auto",
+                    objectFit: "contain",
+                    transform: `rotate(${rotation * 90}deg)`,
+                    transformOrigin: "center",
+                    display: "block",
+                    margin: "auto",
```
alexfauquette: I still don't understand why you had to modify those lines

Ramneet04: hey! actually previously when I altered the logic somewhere it was breaking and later when I wrapped it differently the properties got irrelevant so yeah I have fixed them now.

PR #1365, opened 2025-12-22
chore(deps): upgrade @openfoodfacts/openfoodfacts-webcomponents

piyushzingade:
### What
-  Upgrade the @openfoodfacts/openfoodfacts-webcomponents version from 1.14.2 to 1.15.1 Issue: openfoodfacts/hunger-games/issues/1266

### Fixes bug(s)
- 

### Part of 
-

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,8 +80,5 @@
       "last 1 safari version"
     ]
   },
-  "resolutions": {
-    "@babel/runtime": "^7.23.3"
-  },
   "packageManager": "yarn@4.9.2"
 }


package.json
```diff
@@ -80,5 +80,8 @@
       "last 1 safari version"
     ]
   },
+  "resolutions": {
+    "@babel/runtime": "^7.23.3"
+  },
```
alexfauquette: Why is this needed?

piyushzingade: Oops, accidentally added that.

PR #1384, opened 2026-01-04
refactor: Split QuestionCard into smaller components

VaiTon:


src/components/QuestionCard.tsx
```diff
@@ -1,5 +1,4 @@
-import * as React from "react";
-
+import { MouseEventHandler, useEffect, useState } from "react";
```
alexfauquette: Here I find it better to import `* as React` because in the file you get `React.MouseEventHandler` which make it cristal clear "This is from react"

src/components/QuestionCard.tsx
```diff
@@ -22,59 +21,185 @@ import logo from "../assets/logo.png";
...
-  title,
-  showFilterResume,
-  editableTitle,
-}: QuestionCardProps) => {
-  const { t } = useTranslation();
-
-  const targetUrl = `/questions?${getQuestionSearchParams(filterState)}`;
-
-  const [questionNumber, setQuestionNumber] = React.useState("?");
+const useQuestionCount = (filterState: FilterState) => {
```
alexfauquette: Could be nice to use `useQuery()` to simplify this hook

src/components/QuestionCard.tsx
```diff
@@ -22,59 +21,185 @@ import logo from "../assets/logo.png";
...
     setIsEditMode(false);
   };
-  const stopEdition = () => {
+
+  const handleCancel = () => {
     setInnerTitle(title);
     setIsEditMode(false);
   };
+
+  const handleKeyDown = (event: React.KeyboardEvent) => {
```
alexfauquette: Not sure about where this `React` come from :)

src/components/QuestionCard.tsx
```diff
@@ -22,59 +21,185 @@ import logo from "../assets/logo.png";
...
+  const [isEditMode, setIsEditMode] = useState(false);
+  const [innerTitle, setInnerTitle] = useState(title);
+
+  useEffect(() => {
     setInnerTitle(title);
   }, [title]);
 
-  const startEdition = () => {
+  const handleStartEdit: MouseEventHandler = (event) => {
+    event.stopPropagation();
```
alexfauquette: If it's because the card is a link, the `event.preventDefault()` is better.

`stopPropagation` is something to avoid as much as possible, because it stops the event propagation, so you can do nothing form parents

src/components/QuestionCard.tsx
alexfauquette: I get why, but in the same file modification you replaced 

```diff
- import * as React from "react";
+ import { MouseEventHandler, useEffect, useState } from "react"; 
```

So `React` is not defined anymore

PR #1385, opened 2026-01-04
refactor: Convert SmallQuestionCard to TypeScript

VaiTon:
Add prop types and change questionNumber to number | null. Show a CircularProgress in the badge while loading and replace React.useState/Effect with named imports; consolidate MUI imports.

### DIFF
--- a/src/components/SmallQuestionCard.tsx
+++ b/src/components/SmallQuestionCard.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from "react";
+import * as React from "react";
 
 import {
   CircularProgress,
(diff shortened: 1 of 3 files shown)


src/components/SmallQuestionCard.tsx
```diff
@@ -1,23 +1,41 @@
-import * as React from "react";
+import { useEffect, useState } from "react";
```
alexfauquette: I don't really see why doing this modification

src/components/SmallQuestionCard.tsx
```diff
@@ -1,23 +1,41 @@
...
+type SmallQuestionCardProps = {
+  filterState: Partial<{
+    insightType: string;
+    brandFilter: string;
+    valueTag: string;
+    countryFilter: string;
+    sortByPopularity: boolean;
+    campaign: string;
+    predictor: string;
+    with_image?: boolean;
```
alexfauquette: Might be more interesting to move this to a dedicated type in robotoff.ts to share this filter state type with other files

alexfauquette (approved):

PR #1499, opened 2026-05-24
fix(questions): hide filter chip when URL value is unresolvable

### LINKED ISSUE #1360: Auto-remove undefined when arriving with malformed permalink
### What
Auto-remove undefined when arriving with malformed permalink
https://world.openfoodfacts.org/facets/categories/apple-juices 
>>
https://hunger.openfoodfacts.org/questions?country=en%3Aworld&type=category&value_tag=en%3Aapple-juices
>>

Image

sarthikadhawan:
### What
Filter chips for country, campaign, and predictor used to render "<key>: undefined" when the URL value didn't resolve via lookup, e.g. external permalinks like `?country=en:world` or string-concatenated JS `undefined` from another OFF page.

Gate each chip's display on whether the URL value matches an option the corresponding dropdown would produce (new `isKnownFilterValue` helper). The URL itself is preserved verbatim, only the chip rendering changes, so back/forward navigation and external permalinks keep working unchanged.

Also dedupes the predictor dropdown options into a shared array in `const.ts` so FilterDialog and the new validator share one source of truth.

### Screenshot

| Before | After |
|:-----------------:|:----------------:|
|Fixed issueimage|Screenshot 2026-05-24 185159 |
|Sanity pass image|image|

### Fixes bug(s)
Fixes #1360

### DIFF
--- a/src/components/QuestionFilter/const.ts
+++ b/src/components/QuestionFilter/const.ts
@@ -18,6 +18,21 @@ export const DEFAULT_FILTER_STATE: FilterState = {
 
 export const campagnes = ["agribalyse-category"];
 
+// Drives both the FilterDialog dropdown and the chip-display check in
+// QuestionFilter. `labelKey` resolves under the i18n namespace
+// `questions.filters.predictor.<labelKey>`.
+export const predictors = [
+  { value: "ridge_model-ml", labelKey: "ridge_model_ml" },
+  { value: "neural", labelKey: "neural" },
+  { value: "matcher", labelKey: "matcher" },
+  { value: "google-could-vision", labelKey: "google_cloud_vision" },
+  { value: "regex", labelKey: "regex" },
+  { value: "flashtext", labelKey: "flashtext" },
+  { value: "nutriscore", labelKey: "nutriscore" },
+  { value: "universal-logo-detector", labelKey: "universal_logo_detector" },
+  { value: "ocr", labelKey: "ocr" },
+];
+
 export const countryNames = [
   "",
   "en:australia",
--- a/src/pages/questions/FilterDialog.tsx
+++ b/src/pages/questions/FilterDialog.tsx
@@ -23,6 +23,7 @@ import countries from "../../assets/countries.json";
 import {
   insightTypesNames,
   campagnes,
+  predictors,
 } from "../../components/QuestionFilter/const";
 import { useFilterState } from "../../hooks/useFilterState";
 import { BrandFilter } from "../../components/QuestionFilter/BrandFilter";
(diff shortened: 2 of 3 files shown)


src/components/QuestionFilter/const.ts
```diff
@@ -18,6 +20,44 @@ export const DEFAULT_FILTER_STATE: FilterState = {
...
+  { value: "neural", labelKey: "neural" },
+  { value: "matcher", labelKey: "matcher" },
+  { value: "google-could-vision", labelKey: "google_cloud_vision" },
+  { value: "regex", labelKey: "regex" },
+  { value: "flashtext", labelKey: "flashtext" },
+  { value: "nutriscore", labelKey: "nutriscore" },
+  { value: "universal-logo-detector", labelKey: "universal_logo_detector" },
+  { value: "ocr", labelKey: "ocr" },
+];
+
```
VaiTon, line 35: Is this part of this PR?

sarthikadhawan: yes, moved the predictors list from FilterDialog to const file to be reused in validating URL

src/components/QuestionFilter/const.ts
```diff
@@ -18,6 +20,44 @@ export const DEFAULT_FILTER_STATE: FilterState = {
...
+ * the dropdown would produce. The chip in QuestionFilter is shown only
+ * when this returns true — values that don't match (including the literal
+ * string "undefined" from malformed permalinks) are still kept in the
+ * URL, but no chip is rendered for them. Issue #1360.
+ */
+export const isKnownFilterValue: Record<string, (v: string) => boolean> = {
+  country: (v) => validCountryCodes.has(v),
+  campaign: (v) => validCampaigns.has(v),
+  predictor: (v) => validPredictors.has(v),
+};
```
VaiTon: I don't think the value added by creating these sets and function is enough. We can just `.contains` on the arrays.

src/hooks/useUrlParams.js
```diff
@@ -13,7 +13,11 @@ export const getDefaultizedUrlParams = (defaultParameters, synonyms = {}) => {
   const urlParams = new URLSearchParams(window.location.search);
   Object.keys(defaultParameters).forEach((key) => {
     const value = urlParams.get(key);
-    if (value !== null && JSON.stringify(defaultParameters[key]) !== value) {
+    if (
+      value !== null &&
+      value !== "undefined" &&
+      JSON.stringify(defaultParameters[key]) !== value
+    ) {
```
VaiTon: I would be more interested in understanding how a literal `"undefined"` arrives here. Defenses like these are just adding cognitive complexity to the code IMHO.

src/hooks/useUrlParams.js
```diff
@@ -50,11 +55,20 @@ export const convertObjectParamsToUrlParams = (
...
+    const value = parameters[key];
+    // Skip empty / nullish / literal-"undefined" values so we don't emit
+    // links like `?country=undefined` when a caller forwards a JS `undefined`.
     if (
-      JSON.stringify(parameters[key]) !== urlParams.get(key) &&
-      parameters[key] !== ""
+      value === undefined ||
+      value === null ||
+      value === "" ||
+      value === "undefined"
```
VaiTon: A simple `value == null || value === ''` should do the trick here.

VaiTon (changes requested):

src/hooks/useUrlParams.js
alexfauquette: Or `!value` :)

src/hooks/useUrlParams.js
sarthikadhawan: updated

alexfauquette (approved):
Look nice :+1:

Issue #1514, opened 2026-06-16
paper cuts


teolemon:
### Story

- [ ] Make country and language settings filterable (easier to find your country easily) <img width="648" height="390" alt="Image" src="https://github.com/user-attachments/assets/74fcf039-d299-4c88-86d7-fe18aa1cc9ba" />
- [ ] Make this link into a button to make it easily tappable on mobile <img width="322" height="79" alt="Image" src="https://github.com/user-attachments/assets/b5ff8308-10a3-4404-bce6-3c3dfb8e9bca" />
- [ ] Remove the outdated call to donation, and use the webcomponent for an always up to date banner: "We need your donations to fund the Open Food Facts 2024 budget and to continue to develop the project."
- [ ] Make the top right country menu filterable <img width="506" height="735" alt="Image" src="https://github.com/user-attachments/assets/59bbf2e1-ee6a-43c8-bb96-28930b4a73d4" />
- [ ] Remove or fix the top right question mark in navigation. It looks out of place, and the glyph with a circle around it is more commonly used
- [ ] Revamp the footer
- [ ] Use the App install webcomponent
- [ ] Add the new social networks of Open Food Facts from: https://world.openfoodfacts.org/follow-open-food-facts <img width="506" height="179" alt="Image" src="https://github.com/user-attachments/assets/705d9c4c-3d6e-415b-8639-179080849661" />

ganga-300:
@teolemon I would like to work on this issue.
please assign me

Animesh-Jantwal:
Hi @teolemon , I understand this issue and I would like to work on it.

The current UI has a few parts that are not very user-friendly, especially on mobile. Some links are hard to tap, the country selection is not easy to filter, and some older elements in the header/footer should be updated to match the current design.

My plan is to improve these parts in a simple and clear way:

make the country and language selection easier to find and use
make the link easier to tap on mobile by turning it into a button
replace the outdated donation content with the updated banner
fix or remove the top-right question mark icon so it looks more consistent
refresh the footer and add the updated social links
use the app install webcomponent where appropriate
The goal is to make the interface easier to use, more modern, and more consistent for users.

teolemon:
sure @Animesh-Jantwal 
one papercut, more or less one PR please, easier to review that way

Animesh-Jantwal:
Sure i am working on it now @teolemon

PR #1529, opened 2026-07-06
fix: address Open Food Facts UI paper cuts (filterable selects, webco…

Animesh-Jantwal:
- Addresses the paper-cuts issue #1514.

Changes

1. Filterable country and language in Settings — swapped the plain dropdowns for searchable Autocomplete fields.
2. Report-an-issue link is now a button — a bigger, easier tap target on mobile.
3. Donation banner — replaced the outdated hardcoded donation box with the donation-banner webcomponent, so it stays up to date automatically.
4. Filterable header country menu — swapped the header Select for a searchable Autocomplete.
5. Header help icon — replaced the bare question mark with the circled HelpOutline icon and a tooltip.
6. New social networks — added YouTube, TikTok, Threads, Mastodon and Bluesky (from the official follow page), and updated the X/Twitter link.
7. App install — replaced the hardcoded store links with the mobile-badges webcomponent.
8. Footer cleanup — translated the hardcoded "Discover the project" chip and replaced hardcoded hex colors with theme colors, which also fixes dark mode.

Tested locally in both light and dark mode.


alexfauquette:
Not sure the darck mode background color is the best one for this section. a lighter one might better fit

image


Do we want to add all those social netwoks? I assume most of them are dead

image


Couls you split the PR in different ones as explained in the issue. At least one part page impacted. For example

- one for the usage of web-components
- one for the update of the country select
- one for the localization aspect

Otherwise it's too easy to miss a point to review

Animesh-Jantwal:
Thanks for the thorough review @alexfauquette!  All make sense — I'll split this up.

I'll close this PR and open focused ones per area:
1. Web components (donation-banner + mobile-badges app install)
2. Filterable country / language selects
3. Header help icon
4. Footer social networks
5. Footer localization + dark-mode colors

On your other points:

Social networks: these actually come straight from the official [Follow Open Food Facts](https://world.openfoodfacts.org/follow-open-food-facts) page, so they should be the current accounts — but happy to trim to just the main ones. Which would you like to keep (e.g. Instagram, X, Facebook, YouTube)?

Dark-mode background: good catch — I'll switch that section to a lighter shade in the colors PR. Any preferred value from the theme, or shall I pick a lighter one?

Will get the split PRs up shortly!

PR #1530, opened 2026-07-06
fix: load home page user statistics from /facets with credentials

### LINKED ISSUE #1526: Provide credentials when fetching user statistics
User statistics are not displayed anymore on the main page:

Image

First, all facets URLs should now be prefixed by /facets.

Then, the issue comes from the IP-independent rate-limit we added to facets for non-logged in users. We should add credentials information to this request so that it is not blocked.

Animesh-Jantwal:
Problem

The Editor, Contributor, and Photographer statistics on the home page were showing N/A instead of the actual counts.

Cause

The requests that fetch these counts were failing for two reasons:

1. Wrong URL — Facet pages on Open Food Facts are now served under a /facets prefix. The old paths (/editor/..., /contributor/..., /photographer/...) no longer return data.
2. Blocked by rate limit — A new IP-independent rate limit blocks facet requests from non-logged-in users, so the calls were being rejected.

Fix

- Added the /facets prefix to all three requests.
- Added withCredentials: true so each request sends the user's Open Food Facts login cookie and passes the rate limit.

Notes

@alexfauquette also reported a CORS error when sending credentials. That needs the Open Food Facts server to allow credentialed requests from the Hunger Games origin, which is a server-side change outside this PR.

Closes #1526

### DIFF
--- a/src/pages/home/UserData.tsx
+++ b/src/pages/home/UserData.tsx
@@ -13,19 +13,25 @@ import { OFF_URL } from "../../const";
 
 const fetchUserData = async (userName: string) => {
   const editorPromise = axios
-    .get(`${OFF_URL}/editor/${userName}.json?fields=count`)
+    .get(`${OFF_URL}/facets/editor/${userName}.json?fields=count`, {
+      withCredentials: true,
+    })
     .then(({ data }) => {
       return data?.count;
     })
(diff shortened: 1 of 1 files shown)


alexfauquette:
Already fixed by #1528

PR #1575, opened 2026-08-01
fix: accept world as a valid filter and do not filter as a result

### LINKED ISSUE #1574: The question game is not able to accept en:world as a parameter
- When you come from the world version of open food fact from the facets, the link built contains en:world 
- The question game is not able to accept en:world as a parameter
- As a result, the game errors and the first impression of people is a game that's not working.

teolemon:
### What
- fix: accept world as a valid filter and do not filter as a result


teolemon:
lol, copilot qui casse le thermomètre…

Issue #1558, opened 2026-07-30
Generate a daily/hourly cache of the Categories count for the Green-Score game


teolemon:
### Problem
- Generate a daily cache of the Categories count for the Green-Score game. We should generate one for each country. Potentially a github action to generate and comit the daily cache ?
- do not block the page loading on the count
- Display all categories from cache without the load more button
- Allow to filter the DOM to find a category
- Allow to rank by alphabetical order, or by count

That should be good enough, and increase playability

Animesh-Jantwal:
working on it

PR #1594, opened 2026-08-24
feat: show all Green-Score categories with search and sorting

Animesh-Jantwal:
Refs #1558

Addresses points 2-5 of the issue. The daily cache and GitHub Action (point 1)
are deliberately left out — see the note at the bottom.

## The problem

The category list loads 25 at a time behind a "Load more" button. I measured
the Robotoff endpoint:

| request | rows | payload | time |
|---|---|---|---|
| `count=25` (current) | 25 | 629 B | 4.59 s |
| `count=1000` | 937 | 25 KB | 4.94 s |

**The response time is the query, not the transfer.** Asking for everything
costs the same as asking for 25.

So seeing all 937 categories today means clicking "Load more" 38 times and
waiting ~5s each — about 3 minutes — to transfer 25 KB that arrives in one
request.

## Changes

All in `src/components/Opportunities.tsx`, plus 5 new `green-score.*` keys.

- **One request, no pagination.** Removed the `page` state and the "Load more"
  button. Also removed the now-redundant reset effect, since the list is
  replaced rather than appended to.
- **Search box.** Matches the translated name *and* the raw tag — category
  names arrive from a second async call, so tag-matching keeps search working
  in the meantime.
- **Sort selector.** Number of questions (default) or name A-Z.
- **Controls render immediately** rather than behind the loading state.

## Batching the taxonomy lookups

Removing pagination surfaced a problem worth calling out. `getCategoriesTranslations`
passes every tag in the query string:

| tags | URL length | result |
|---|---|---|
| 400 | 6,803 | 200 OK |
| 500 | 8,930 | **414 URI Too Long** |
| 937 | 18,933 | **connection dropped** |

Fine at 25 per page. With all 937 it fails outright, and every card falls back
to showing its raw tag (`en:chickens` instead of "Chickens"). Lookups are now
batched at 100 tags per request, with a ref-tracked set so no tag is requested
twice.

## Incidental fixes

- **Typed the component.** Error count was climbing against an already-messy
  file, so I added proper types. `eslint src/components/Opportunities.tsx` now
  reports **0 errors, down from 29** on master.
- **Race condition:** the fetch `.catch` didn't check `isValid`, so a quickly
  abandoned request could clear the loading state of the one that replaced it.
- Replaced the hand-written `[0, 1, 2, ... 24]` skeleton array with
  `Array.from({ length: 12 })`.

## On the daily cache (point 1)

I left it out on purpose and would like your steer before building it:

- 252 countries × ~5 s per request ≈ **21 minutes** of hammering Robotoff daily
- A daily commit touching hundreds of files is significant repo churn
- Realistically only ~50-100 countries have meaningful data

Would you prefer caching `world` plus a top-N set of countries, with a live API
fallback for the rest? Happy to do it as a follow-up PR once the shape is agreed.

## Verification

Verified locally:

- `npm run build` passes
- `npx eslint src/components/Opportunities.tsx` — clean, 0 errors (29 on master)
- `prettier --check` clean on all three files
- New i18n keys confirmed present exactly once in `en.json` and `common.json`
- API measurements above taken directly against the live Robotoff and
  taxonomy endpoints

**Not yet verified in a browser** — I'd appreciate a reviewer confirming the
search/sort interaction and that category names resolve correctly with the
batched lookups.

### DIFF
--- a/src/components/Opportunities.tsx
+++ b/src/components/Opportunities.tsx
@@ -1,5 +1,6 @@
 import * as React from "react";
 import { Link } from "react-router";
+import { useTranslation } from "react-i18next";
 
 import Typography from "@mui/material/Typography";
 import Card from "@mui/material/Card";
@@ -7,7 +8,9 @@ import CardActionArea from "@mui/material/CardActionArea";
 import CardContent from "@mui/material/CardContent";
 import Box from "@mui/material/Box";
 import Skeleton from "@mui/material/Skeleton";
-import Button from "@mui/material/Button";
+import Stack from "@mui/material/Stack";
+import TextField from "@mui/material/TextField";
+import MenuItem from "@mui/material/MenuItem";
 
 import Loader from "../pages/loader";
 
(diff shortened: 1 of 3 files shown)


teolemon:
@Animesh-Jantwal I have made a similar vibecoded PR: https://github.com/openfoodfacts/hunger-games/pull/1584
It's not working very well, but you can see my intent there.

Animesh-Jantwal:
okay noted @teolemon

PR #1598, opened 2026-08-27
feat(questions): make missing product fields actionable

### LINKED ISSUE #1545: Make the ??? actionable to add missing product information in the Questions game
- Make the ??? actionable
- We could start by a link to open edit mode for the product.
- We could inline robotoff predictions for that specific field, similar to what's in the app (it could only be a chip: (Lasagne (v)(x))

Image

Image

Animesh-Jantwal:
Right now, when a product is missing a field, the info panel in the Questions game just shows a plain `?`. It's a dead end — you're looking straight at a photo of the packaging, you can see the missing information, but there's no way to act on it without hunting for the Edit button and starting over.

This implements the first step you suggested: each missing row now shows an **Add** link that opens the product's edit page, deep-linked to the relevant field with an anchor.

### What changed

- `ADDITIONAL_INFO_TRANSLATION` gains an optional `editAnchor` per field
- Missing rows render an `Add` link instead of the dead `?`
- One new translation key, `questions.add_missing_info`, in `common.json`

### Also fixed

The old fallback was `value ?? "?"`, which only catches `null` and `undefined`. A field that came back as an **empty array** slipped past it, hit the `Array.isArray` branch, mapped over nothing and rendered a completely blank cell — not even a `?`. The new `isMissing` check treats empty strings and empty arrays as missing too.

### Worth confirming

I couldn't verify the `#brands` / `#categories` style anchors myself — `product.pl?type=edit` requires login, so logged out I only get the sign-in form. If the real field ids differ, the `editAnchor` values need adjusting. The code degrades safely either way: no anchor just opens the top of the edit page, still better than a dead `?`.

### Follow-up

I've deliberately left the second half of the issue — inline Robotoff prediction chips per field — for a separate PR. The plumbing already exists (`useProductQuestions` + the Yes/No/Send pattern in `ProductOtherQuestions`), and insights carry an `insight_type` that maps onto four of these six rows, so it's very doable as a focused follow-up.

### Question

For the missing rows, would you prefer this `Add` link, or keeping the `?` and making it clickable? The link is more discoverable but slightly noisier when a product is missing several fields at once. Happy to switch.

Part of #1545

### DIFF
--- a/src/i18n/common.json
+++ b/src/i18n/common.json
@@ -91,6 +91,7 @@
     "no_similar_questions": "No similar questions",
     "no_images": "No images available",
     "no_additional_info": "No additional information",
+    "add_missing_info": "Add",
     "filters": {
       "active_filter_number": "+{{count}} filters",
       "active_filter_number_one": "+{{count}} filter",
--- a/src/i18n/en.json
+++ b/src/i18n/en.json
@@ -92,6 +92,7 @@
     "no_similar_questions": "No similar questions",
     "no_images": "No images available",
     "no_additional_info": "No additional information",
+    "add_missing_info": "Add",
     "filters": {
       "active_filter_number": "+{{count}} filters",
       "active_filter_number_one": "+{{count}} filter",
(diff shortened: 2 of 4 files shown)


teolemon:
@Animesh-Jantwal missing strings
image

Animesh-Jantwal:
@teolemon 

Good catch, thanks — fixed in 50c61b63.

I'd added the key to `common.json` but not `en.json`. Since `i18n.ts` loads
`src/i18n/<lang>.json` at runtime and `common.json` is only the Crowdin
source, the key never existed at runtime and i18next fell back to printing
the path.

It now reads "Add" as intended.

teolemon (approved):
Thanks @Animesh-Jantwal 👍

PR #1603, opened 2026-08-29
fix(favorites): fall back to the logo when a saved image is broken

### LINKED ISSUE #854: Saved queries point to a broken asset
### What
- Saved queries point to a broken asset
- https://hunger.openfoodfacts.org/static/media/logo.65a86b94902024e8f468.png
- Probably is this: https://github.com/openfoodfacts/hunger-games/blob/master/src/assets/logo.png

image

Animesh-Jantwal:
### What was happening

Old favourites saved the **build-time URL** of `logo.png` into localStorage. Back then the app used Create React App, so that URL looked like:

`/static/media/logo.65a86b94902024e8f468.png`

After the move to Vite, that path stopped existing. It still returns 404 today, so anyone with an old saved filter sees a broken image on the home page — exactly your screenshot.

### Good news: new saves are already fine

The code that caused this is no longer in use. Saving a favourite now goes through `useFavorite`, which stores an empty string instead of a URL, and the card falls back to the bundled logo. So this only affects favourites saved before that change — but for those users it never heals, because the bad URL just sits in their browser.

### The fix

If the saved image fails to load, the card now falls back to the bundled logo. That repairs old favourites without needing to touch anyone's stored data, and also covers remote images that disappear later.

The handler only runs once per image, so there's no loop if the fallback itself ever fails.

### Two related things I left alone

**The buggy code still exists, unused.** `useFilterSearch` in `src/components/QuestionFilter/useFilterSearch.js` still does `localFavorites.addQuestion(exposedParameters, logo, title)`, which is the exact line that caused this. Nothing imports it — `knip` lists it under "Unused exports" — but if someone wired it up, the bug would come straight back. Worth deleting; happy to do it in a separate PR.

**`SmallQuestionCard`** renders images the same way and could show a broken image if a remote logo goes missing. Not part of this issue, but the same fallback would work there.

Fixes #854

## Summary by CodeRabbit

- **Bug Fixes**
  - Added a fallback logo when question card images fail to load.
  - Prevented repeated fallback attempts that could cause image-loading loops.


teolemon:
@coderabbitai full review

VaiTon (approved):

PR #1619, opened 2026-08-29
fix: bump deps

teolemon:
### What
- fix: bump deps to solve security reports

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1050,6 +1050,15 @@ __metadata:
   languageName: node
   linkType: hard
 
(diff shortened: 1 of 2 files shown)


VaiTon (approved):

yarn.lock
```diff
@@ -7337,6 +7320,13 @@ __metadata:
   languageName: node
   linkType: hard
 
+"yallist@npm:^5.0.0":
```
teolemon, line 7323: why is it adding that ?

PR #1534, opened 2026-07-07
feat: use donation-banner and mobile-badges webcomponents in footer

Animesh-Jantwal:
Part of the paper-cuts issue #1514 — splitting #1529 into focused PRs, per @alexfauquette's review.

Replaces two hardcoded footer sections with official Open Food Facts webcomponents so they stay up to date automatically:

1. Donation box → `<donation-banner>` webcomponent
2. App install store links → `<mobile-badges>` webcomponent

Also removes the now-unused `FooterButtons` component and adds JSX type declarations for the two custom elements.

PR #1674, opened 2026-09-07
fix(footer): drop the light backdrop behind the app badges

### LINKED ISSUE #1668: The app install banner has an ugly white line around it in dark mode
### Story
- The app install banner has an ugly white line around it in dark mode
- Unsure if already fixed in openfoodfacts-webcomponents, if so we need to update our version of openfoodfacts-webcomponents

Image

Animesh-Jantwal:
### It wasn't the web component

You wondered whether this was already fixed upstream. It isn't an upstream bug — the white line comes from our own code, and I added it, in #1534. Sorry about that.

`DownloadOpenFoodFacts.jsx` painted a light `#f5f5f5` panel behind the banner in dark mode, with padding around it. The padding is the pale border in your screenshot.

For what it's worth, `1.16.0` is already the latest published version of openfoodfacts-webcomponents, so there's no update to pick up here.

### Why the panel was there

`mobile-badges` styles itself for dark mode — it sets its own `background-color: #201a17` and light text. So the backdrop is redundant most of the time.

But it decides that from **`window.matchMedia("(prefers-color-scheme: dark)")`**, the operating system setting — not from our own light/dark toggle. So when someone's system is light but they've switched Hunger Games to dark, the banner renders in its light style inside a dark page and the text becomes hard to read. That's the case the panel was covering.

### The fix

The backdrop now applies only when those two actually disagree:

```js
const needsLightBackdrop = theme.palette.mode === "dark" && !systemPrefersDark;
```

- System dark + app dark → no backdrop, component's own dark styling shows, **white line gone**
- System light + app dark → backdrop kept, badges stay readable
- App in light mode → unchanged

Your screenshot is the first case, so the border should be gone for you.

### Worth doing upstream

The real fix is for `mobile-badges` to accept a dark-mode override so a host app can tell it which theme it's using, instead of it reading the OS directly. Then this workaround could go away entirely. Happy to open an issue on openfoodfacts-webcomponents for that if you think it's worth it.

### Verification

- `prettier --check` — clean
- `eslint` — no findings in this file
- `tsc --noEmit` — clean
- `yarn build` — passes

Fixes #1668

Screenshot 2026-09-07 at 4 22 18 PM

## Summary by CodeRabbit

- **Style**
  - Improved the download badge’s appearance for dark themes when the system uses a light color scheme.
  - Added a light backdrop, rounded corners, and spacing in this scenario.
  - Preserved the transparent, unpadded appearance when dark mode is preferred by the system.

### DIFF
--- a/src/components/Footer/DownloadOpenFoodFacts.jsx
+++ b/src/components/Footer/DownloadOpenFoodFacts.jsx
@@ -1,6 +1,11 @@
-import { Box } from "@mui/material";
+import { Box, useMediaQuery, useTheme } from "@mui/material";
 
 export default function DownloadOpenFoodFacts() {
+  const theme = useTheme();
+  const systemPrefersDark = useMediaQuery("(prefers-color-scheme: dark)");
+  const needsLightBackdrop =
+    theme.palette.mode === "dark" && !systemPrefersDark;
+
   return (
     <Box
       className="OFF-download"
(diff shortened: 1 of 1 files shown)


teolemon (approved):
Thanks for the investigation @Animesh-Jantwal 👍

Animesh-Jantwal:
My pleasure @teolemon

Issue #1671, opened 2026-09-04
Add a promo for new product creation in Open Prices


teolemon:
### Story
> [!WARNING]  
> This issue is not ready to be picked up. It needs agreement, a mockup and a technical solution first !

- Add a promo for new product creation in Open Prices
- https://prices.openfoodfacts.org/experiments/create-off-product
- Create missing products

sarthaklole:
Hi @teolemon  
I'm interested in picking this up and wanted to help move it from "To Discuss & Validate" forward.

Before proposing a mockup/technical approach, I have a question on the trigger mechanism:

Since this promo needs to detect a "missing/incomplete product" state to know when to show itself — is there already a shared utility or API response flag in Hunger Games that identifies this case, or would this be the first feature needing that check?

I looked through the repo and didn't see an existing helper for this specifically — if I missed it, please point me to it. If it doesn't exist yet, I'd like to propose building it as a reusable check so future features (not just this promo) can rely on the same detection logic instead of duplicating it.

raphodn:
that's cool, thanks :)
and I see there's a similar issue: #1670

teolemon:
fixed

PR #1677, opened 2026-09-07
feat(home): promote product creation on Open Prices

Animesh-Jantwal:
Adds a fourth card to the home selector pointing at the Open Prices create-off-product experiment.

I know this is marked as needing agreement and a mockup first — I've built it as something concrete to react to rather than a finished proposal. Happy to change or close it.

### What's here

- A **Create missing products** card next to Questions, Logos and Green-Score
- Links to https://prices.openfoodfacts.org/experiments/create-off-product, opening in a new tab
- Existing cards are untouched; the card definition just gained an optional `href`, and external ones render as a link instead of a router route

The reasoning for the placement: someone who has run out of questions to answer is exactly the person who might go and add the products that are missing altogether. The game selector is where they're already looking.

### The image is a placeholder

This is the part that most needs your input. Open Prices doesn't publish a logo I could use — I checked, and there's only a `favicon.ico`, which is far too small for a 200px card. So the card currently shows an MUI shopping-cart icon.

If there's a proper Open Prices asset somewhere, point me at it and I'll swap it in. Otherwise this could use an illustration in the same style as the new Questions and Logos SVGs.

### Wording

Also placeholder, and easy to change — it's two new translation keys:

- **Create missing products**
- *Add products that are not in the database yet, over on Open Prices*

### Verification

- Target URL returns 200
- `prettier --check` — clean
- `eslint` — no findings
- `yarn build` — passes
- Keys added to both `common.json` and `en.json`

Refs #1671

### DIFF
--- a/src/i18n/common.json
+++ b/src/i18n/common.json
@@ -14,6 +14,10 @@
         "green_score": {
           "title": "Green-score",
           "description": "Annotate the questions with the highest value"
+        },
+        "create_products": {
+          "title": "Create missing products",
+          "description": "Add products that are not in the database yet, over on Open Prices"
         }
       }
     },
--- a/src/i18n/en.json
+++ b/src/i18n/en.json
@@ -14,6 +14,10 @@
         "green_score": {
           "title": "Green-score",
           "description": "Annotate the questions with the highest value"
+        },
+        "create_products": {
+          "title": "Create missing products",
+          "description": "Add products that are not in the database yet, over on Open Prices"
         }
       }
     },
--- a/src/pages/home/homeCards.jsx
+++ b/src/pages/home/homeCards.jsx
@@ -70,10 +70,12 @@ const HomeCards = () => {
                   display: "flex",
                   alignItems: "center",
                   justifyContent: "center",
+                  backgroundColor: "#f6f3f0",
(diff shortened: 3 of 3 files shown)


teolemon:
- I'm fine with the placeholder, let's decide we can use an icon like that with the theme color when we don't have a bespoke visual.
- Also, I'm actually fine with increasing the number of games on the homepage. That's touch friendly on mobile, and also works on desktop.
- Can you adjust the radius to be the same as the other cards @Animesh-Jantwal ?

Animesh-Jantwal:
Done — the panel is now 16px to match.

I checked the card SVGs rather than eyeballing it: both draw their background
as `<rect width="350" height="200" rx="16">` with a `0 0 350 200` viewBox, so
16px is exactly what the illustrated cards use. Mine was 8px.

Good to hear on the placeholder convention — the icon uses the OFF palette
colours already in the theme (latte panel, mocha icon), so it should stay
consistent if we reuse the pattern for other cards without bespoke artwork.

teolemon (approved):
Great @Animesh-Jantwal 👍

Animesh-Jantwal:
My pleasure @teolemon

PR #1682, opened 2026-09-08
fix(cards): give logo cards the same dark mode treatment everywhere

### LINKED ISSUE #1679: Card harmonization through Hunger Games
### Story

Image

Image

Image

Animesh-Jantwal:
Starts on the harmonisation, with the part that's an objective inconsistency rather than a design choice.

### The problem

Most label logos are black artwork on a transparent background, so they vanish against a dark surface. #1601 fixed that on the Logos Dashboard by sitting each logo on a light panel in dark mode — but only there.

Three components render logos this way, and only one had the fix:

| Component | Used by | Before |
|---|---|---|
| `DashboardCard` | Logos Dashboard | had the panel |
| `SmallQuestionCard` | Green-Score | no panel |
| `QuestionCard` | Saved filters | no panel |

That's your second screenshot: on the Green-Score page the Fairtrade, RSPO, Rainforest Alliance and MSC logos were sinking into the background, while the same logos looked fine on the dashboard.

Both now use the treatment the dashboard already had, so a logo looks the same wherever it appears.

### What I have not touched

The issue is three screenshots without a written target, so I've stopped at the part that is unambiguously wrong. The remaining differences are design decisions I did not want to guess at:

- **Image heights** — 150px on the dashboard and Green-Score, 200px on saved filters, 300px cards on the home page
- **Structure** — the dashboard has Questions/Search buttons, Green-Score cards have no title at all, home cards have a title and description
- **Badges** — different positions and colours between pages

Tell me what harmonised should look like for those and I'm happy to do it. My instinct is that the Green-Score and dashboard cards should converge first since they show the same kind of content, and the home cards can stay their own thing since they're navigation rather than data. But that's your call.

### Verification

- `prettier --check` — clean
- `eslint` — no findings in either file
- `tsc --noEmit` — clean
- `yarn build` — passes

Refs #1679

Screenshot 2026-09-08 at 4 15 20 PM

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved image visibility across question cards and dashboard cards by applying a white background and padding in all theme modes.
  * Preserved contained image scaling to prevent distortion across card layouts.


src/components/QuestionCard.tsx
```diff
@@ -255,7 +255,13 @@ export default function QuestionCard({
...
             image={imageSrc || logo}
             alt=""
-            sx={{ objectFit: "contain" }}
+            sx={(theme) => ({
+              objectFit: "contain",
+              ...(theme.palette.mode === "dark" && {
+                backgroundColor: theme.palette.common.white,
+                p: 1,
+              }),
+            })}
```
alexfauquette: Could it be simplifed as follow? Since the goal is to get the same color regardless of the theme variant?

```suggestion
            sx={{
              objectFit: "contain",
                backgroundColor: "white",
                p: 1,
              
            }}
```

teolemon:
Can we go a step further and get the same card structure ? @Animesh-Jantwal @alexfauquette

Animesh-Jantwal:
@alexfauquette good catch — simplified, and I applied the same to
`DashboardCard` so all three stay identical rather than drifting apart again.

@teolemon on the structure: happy to. The blocker is data rather than layout.
`SmallQuestionCard` doesn't take a title at all — the Green-Score cards pass
`title: "en:organic"`, a raw tag used only as a React key, while the dashboard
has proper labels like "Bio".

So my proposal would be:

1. Give the Green-Score cards human labels (most already exist in
   `dashboardDefinition.ts` for the same tags)
2. Have `SmallQuestionCard` render a title, so it matches the dashboard
3. Align the image heights (150 vs 200)

Should the Green-Score cards also get the Questions / Search buttons the
dashboard has, or stay click-through-only? That's the main open question —
the rest I can just do.

